PR #125, opened 2015-01-13
Remove Endpoint and Conn as arguments to Path Helpers

drepo:
It seems giving Endpoint or Conn as first argument to Path Helpers do not work any longer and raise "no function clause matching" error. Have looked up the routing docs on the project and they seem to have updated the examples with only one argument being the action.

### DIFF
--- a/C_routing.md
+++ b/C_routing.md
@@ -158,21 +158,21 @@ Path helpers are functions which are dynamically defined on the `Router.Helpers`
 Their names are derived from the name of the controller used in the route definition. Our controller is `HelloPhoenix.PageController`, and `page_path` is the function which will return the path to the root of our application.
 
 That's a mouthful. Let's see it in action. Run `$ iex -S mix` at the root of the
-project. When we call the `page_path` function on our router helpers with the
-the `Endpoint` or connection and action as arguments, it returns the path to us.
+project. When we call the `page_path` function on our router helpers with action as argument, 
+it returns the path to us.
 
 ```elixir
-iex(4)> HelloPhoenix.Router.Helpers.page_path(Endpoint, :index)
+iex(4)> HelloPhoenix.Router.Helpers.page_path(:index)
 "/"
 ```
 
 This is significant because we can use the `page_path` function in a template to link to the root of our application.
 
 ```html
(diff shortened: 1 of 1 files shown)


lancehalvorsen:
@drepo thanks for submitting this. Are you working from Phoenix v0.8.0 or master? For now, the master branch of the guides is intended to match Phoenix v0.8.0.

In either case, though, I don't believe this is accurate. If we look at the quote block that creates these helpers in master:
https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/router/helpers.ex#L93#L118

We see lines that look like this `def unquote(:"#{helper}_path")(conn_or_endpoint, unquote(action), unquote_splicing(vars)) do` which clearly take `conn_or_endpoint` first.

The proof is in the pudding, though. Here's a new 0.8.0 app.
```console
$ iex -S mix
Erlang/OTP 17 [erts-6.0] [source-07b8f44] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Interactive Elixir (1.0.2) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> HelloPhoenix.Router.Helpers.page_path(Endpoint, :index)
"/"
iex(2)> HelloPhoenix.Router.Helpers.page_path(:index)
** (UndefinedFunctionError) undefined function: HelloPhoenix.Router.Helpers.page_path/1
    (hello_phoenix) HelloPhoenix.Router.Helpers.page_path(:index)
```

drepo:
I am running Phoenix v0.7.2. And I get the opposite on iex. 

Infact, the working has changed between v0.7.2 (http://www.phoenixframework.org/v0.8.0/docs/routing#path-helpers) and v0.8.0 (http://www.phoenixframework.org/v0.8.0/docs/routing#path-helpers).

Anyways, Thanks for taking time to review this. I will close.

```elixir
iex(1)> HelloPhoenix.Router.Helpers.page_path(Endpoint, :index)      
** (FunctionClauseError) no function clause matching in HelloPhoenix.Router.Helpers.page_path/2
    (hello_phoenix) web/router.ex:1:HelloPhoenix.Router.Helpers.page_path(Endpoint, :index)

iex(1)> HelloPhoenix.Router.Helpers.page_path(:index)  
"/"

```

PR #126, opened 2015-01-13
Router Refurb for 080

lancehalvorsen:
- much re-writing, additions, and subtractions to make the guide match up with Phoenix v0.8.0
- small refurb changes suggested by phoenixframework.org

### DIFF
--- a/C_routing.md
+++ b/C_routing.md
@@ -9,6 +9,7 @@ defmodule HelloPhoenix.Router do
   pipeline :browser do
     plug :accepts, ~w(html)
     plug :fetch_session
+    plug :fetch_flash
     plug :protect_from_forgery
   end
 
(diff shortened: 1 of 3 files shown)


C_routing.md
```diff
@@ -82,17 +81,25 @@ Let's see how this works. Go to the root of a newly-generated Phoenix applicatio
...
 
-Put this line into your `router.ex` file inside the `scope "/", HelloPhoenix do` block: `resources "users", UserController`. (For this purpose, it doesn't matter that we don't actually have a `HelloPhoenix.UserController`.)
+Let's add a resource to our `router.ex` file like this.
+
+```elixir
+scope "/", HelloPhoenix do
+  pipe_through :browser # Use the default browser stack
+
+  get "/", PageController, :index
+  resources "users", UserController
```
chrismccord: `resources "/users", UserController`

C_routing.md
```diff
@@ -826,62 +799,36 @@ defmodule HelloPhoenix.Router do
...
+Next, we need to define a channel, specifying a topic and associating it with the channel module which will implement its behavior. If we have a channel module called `RoomChannel` and a topic called `lobby`, the code to do this is straightforward.
 
 ```elixir
 defmodule HelloPhoenix.Router do
   use Phoenix.Router
-  use Phoenix.Router.Socket, mount: "/my_socket"
 
-  channel "our_channel_name", HelloPhoenix.OurChannel
+  socket "/ws", HelloPhoenix do
+    channel "lobby", RoomChannel
```
chrismccord: channel "rooms:*", RoomChannel

lancehalvorsen: Interesting. Do all topics need to be wildcards? And do they all need to have a `:` in them?

C_routing.md
lancehalvorsen: Thanks! Will fix and push now.

C_routing.md
chrismccord: No, they can just be normal strings, but the idiom is "topic:subtopic", so a wildcard pattern is the preferred approach since it pretty much mirrors what we were doing prior to 0.8, but we were conflating what a channel and a topic was. Now, the topic is just a string identifier, and we dispatch to channels based on the topic patterns. The new channel docs should help describe the changes https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/channel.ex

"bare" topics make sense for some use-cases, like a "system" topic where you have just one singleton thing, i.e.:

```elixir
channel "system", SystemChannel
...
def join("system", message, socket) ...

def handle_in("set:clock", %{"time" => time})...
```

Maybe. Wildcards offer much more flexibility and is where I'll steer folks towards unless there's a good reason not to.

lancehalvorsen: :+1: Totally makes sense. I'll do a quick fix 'n push right now. Thanks for the explanation!

lancehalvorsen:
@jeregrine @chrismccord I'm going to update phoenixframwork.org with these changes now, just for the record. :ship: :ship: :ship: :shipit:

PR #127, opened 2015-01-14
Updated links from github to phoenixframework.org

drepo:
The Github link no longer has content relevant to the mentioned #links.


drepo:
I think we could also replace Topics in the document with PubSub

lancehalvorsen:
@drepo Thanks for this!

PR #130, opened 2015-01-17
Fix links to phoenixframework.org

lancehalvorsen:
Remove Phoenix version number for links to phoenixframework.org.


lancehalvorsen:
@jeregrine @chrismccord moving these over as well.

PR #134, opened 2015-01-18
Rename destroy to delete

chrismccord:


lancehalvorsen:
Wow, that's been there a while. Thanks @chrismccord!

I'll move these files over to the site now.

chrismccord:
These are breaking changes that I *just* made on Phoenix master so we should leave them off the site until the next version ships. Just doing my fair share :)

lancehalvorsen:
Ah, oops! I was wondering how I got that so wrong for so long. :)

I'll revert the changes on the site. Thanks for your help!

PR #139, opened 2015-02-16
Update C_routing.md

peck:
Appears to be a typo on the topic where lobby is used instead of room


lancehalvorsen:
Thanks for finding this!

PR #146, opened 2015-02-24
Routing update and overall improvements.

kgish:
While reading the section on routing as a (native English), I made a number of corrections and fixed some typos which I believe will improve the overall readability of the text. Hopefully I didn't mess up any of the technical details, but ran the code to double-check so think it's alright.


kgish:
This one isn't correct, sorry.

PR #147, opened 2015-02-24
Routing update and overall improvements.

kgish:
While reading the section on routing as a (native English), I made a number of corrections and fixed some typos which I believe will improve the overall readability of the text. Hopefully I didn't mess up any of the technical details, but ran the code to double-check so think it's alright.

### DIFF
--- a/C_routing.md
+++ b/C_routing.md
@@ -1,4 +1,4 @@
-The router is the main hub of any Phoenix application. It matches HTTP requests to controller actions, wires up realtime channel handlers, and defines a series of pipeline transformations for scoping middleware to sets of routes.
+The router is the main hub of any Phoenix application. It matches HTTP requests to controller actions, wires up real-time channel handlers, and defines a series of pipeline transformations for scoping middleware to sets of routes.
 
 The router file that Phoenix generates, `web/router.ex`, will look something like this one.
 
@@ -36,7 +36,10 @@ The first line of this module `use Phoenix.Router` simply makes Phoenix router f
 Scopes have their own section in this guide, so we won't spend time on the `scope "/", HelloPhoenix do` block here. The `pipe_through :browser` line will get a full treatment in the Pipeline section of this guide. For now, you only need to know that pipelines allow a set of middleware transformations to be applied to different sets of routes.
 
 Inside the scope block, however, we have our first actual route.
-`get "/", PageController, :index`
+
+```elixir
+  get "/", PageController, :index
+```
 
 `get` is a Phoenix macro which expands out to define one clause of the `match/3` function. It corresponds to the HTTP verb GET. Similar macros exist for other HTTP verbs including POST, PUT, PATCH, DELETE, OPTIONS, CONNECT, TRACE and HEAD.
 
@@ -89,7 +92,7 @@ The output tells us that any HTTP GET request for the root of the application wi
 
 The router supports other macros besides those for HTTP verbs like `get`, `post` and `put`. The most important among them is `resources`, which expands out to eight clauses of the match function.
 
-Let's add a resource to our `router.ex` file like this.
+Let's add a resource to our `web/router.ex` file like this.
 
 ```elixir
 scope "/", HelloPhoenix do
@@ -155,17 +158,14 @@ comment_path  GET     /comments/:id/edit             HelloPhoenix.CommentControl
 comment_path  GET     /comments/new                  HelloPhoenix.CommentController.new/2
 comment_path  GET     /comments/:id                  HelloPhoenix.CommentController.show/2
 comment_path  POST    /comments                      HelloPhoenix.CommentController.create/2
-comment_path  PUT     /comments/:id                  HelloPhoenix.CommentController.update/2
 comment_path  PATCH   /comments/:id                  HelloPhoenix.CommentController.update/2
+              PUT     /comments/:id                  HelloPhoenix.CommentController.update/2
 ```
 ###Path Helpers
 
-Path helpers are functions which are dynamically defined on the `Router.Helpers` module for an individual application. For us, that is `HelloPhoenix.Router.Helpers`.
-Their names are derived from the name of the controller used in the route definition. Our controller is `HelloPhoenix.PageController`, and `page_path` is the function which will return the path to the root of our application.
+Path helpers are functions which are dynamically defined on the `Router.Helpers` module for an individual application. For us, that is `HelloPhoenix.Router.Helpers`.  Their names are derived from the name of the controller used in the route definition. Our controller is `HelloPhoenix.PageController`, and `page_path` is the function which will return the path to the root of our application.
 
-That's a mouthful. Let's see it in action. Run `$ iex -S mix` at the root of the
-project. When we call the `page_path` function on our router helpers with the
-the `Endpoint` or connection and action as arguments, it returns the path to us.
+That's a mouthful. Let's see it in action. Run `$ iex -S mix` at the root of the project. When we call the `page_path` function on our router helpers with the the `Endpoint` or connection and action as arguments, it returns the path to us.
 
 ```elixir
 iex(4)> HelloPhoenix.Router.Helpers.page_path(Endpoint, :index)
@@ -177,8 +177,7 @@ This is significant because we can use the `page_path` function in a template to
 ```html
 <a href="<%= HelloPhoenix.Router.Helpers.page_path(@conn, :index) %>">To the Welcome Page!</a>
 ```
-Note: If that function invocation seems uncomfortably long, there is a solution. By including `import HelloPhoenix.Router.Helpers` in our main
-application view, we can shorten that to `page_path(@conn, :index)`. Please see the [View Guide](http://www.phoenixframework.org/docs/views) for more information.
+Note: If that function invocation seems uncomfortably long, there is a solution. By including `import HelloPhoenix.Router.Helpers` in our main application view, we can shorten that to `page_path(@conn, :index)`. Please see the [View Guide](http://www.phoenixframework.org/docs/views) for more information.
 
 This pays off tremendously if we should ever have to change the path of our route in the router. Since the path helpers are built dynamically from the routes, any calls to `page_path` in our templates will still work.
 
@@ -228,14 +227,14 @@ The `Endpoint.url/1` function will get the host, port, proxy port and ssl inform
 
 ###Nested Resources
 
-It is also possible to nest resources in a Phoenix router. Let's say we also have a posts resource which has a one to many relationship with users. That is to say, a user can create many posts, and an individual post belongs to only one user. We can represent that with a nested route like this.
+It is also possible to nest resources in a Phoenix router. Let's say we also have a posts resource which has a one to many relationship with users. That is to say, a user can create many posts, and an individual post belongs to only one user. We can represent that by adding a nested route in `web/router.ex` like this.
 
 ```elixir
 resources "users", UserController do
   resources "posts", PostController
 end
 ```
-When we run `$ mix phoenix.routes` now, in addition to the routes we saw for users above, we get the following set of routes.
+When we run `$ mix phoenix.routes` now, in addition to the routes we saw for users above, we get the following set of routes:
 
 ```elixir
 . . .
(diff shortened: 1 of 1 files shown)


C_routing.md
```diff
@@ -301,58 +300,69 @@ end
 
 Note that Phoenix will assume that the path we set ought to begin with a slash, so `scope "/admin" do` and `scope "admin" do` will both produce the same results.
 
-Note also, that the way this scope is currently defined, we need to fully qualify our controller name, `HelloPhoenix.Admin.ReviewController`. We'll fix that in a minute.
+Note also, that the way this scope is currently defined, we need to fully qualify our controller name, `Admin.ReviewController`. We'll fix that in a minute.
```
lancehalvorsen: Actually, this is not correct. The fully qualified name is "HelloPhoenix.Admin.ReviewController"

C_routing.md
```diff
@@ -301,58 +300,69 @@ end
...
 review_path  POST    /admin/reviews           HelloPhoenix.Admin.ReviewController.create/2
-review_path  PUT     /admin/reviews/:id       HelloPhoenix.Admin.ReviewController.update/2
-             PATCH   /admin/reviews/:id       HelloPhoenix.Admin.ReviewController.update/2
+review_path  PATCH   /admin/reviews/:id       HelloPhoenix.Admin.ReviewController.update/2
+             PUT     /admin/reviews/:id       HelloPhoenix.Admin.ReviewController.update/2
 review_path  DELETE  /admin/reviews/:id       HelloPhoenix.Admin.ReviewController.delete/2
 ```
 
-This looks good, but there is a problem here. Remember that we wanted both user facing reviews routes as well as the admin ones. When we define both of those routes in our router, like this,
+This looks good, but there is a problem here. Remember that we wanted both user facing reviews routes `/reviews` as well as the admin `/admin/reviews` ones. If we now include the user facing reviews in our router like this:
```
lancehalvorsen: I would switch the order of `/admin/reviews` and 'ones' like this: 

. . . as well as the admin ones `/admin/reviews`.

C_routing.md
```diff
@@ -301,58 +300,69 @@ end
...
 review_path  POST    /admin/reviews           HelloPhoenix.Admin.ReviewController.create/2
-review_path  PUT     /admin/reviews/:id       HelloPhoenix.Admin.ReviewController.update/2
-             PATCH   /admin/reviews/:id       HelloPhoenix.Admin.ReviewController.update/2
+review_path  PATCH   /admin/reviews/:id       HelloPhoenix.Admin.ReviewController.update/2
+             PUT     /admin/reviews/:id       HelloPhoenix.Admin.ReviewController.update/2
 review_path  DELETE  /admin/reviews/:id       HelloPhoenix.Admin.ReviewController.delete/2
 ```
 
-The actual routes we get all look right, except for the path helper at the beginning of each line. We are getting the same helper for both the user facing review routes and the admin ones. We can fix this problem by adding an `as: :admin` option to our admin scope.
+The actual routes we get all look right, except for the path helper `review_path` at the beginning of each line. We are getting the same helper for both the user facing review routes and the admin ones which is not correct. We can fix this problem by adding an `as: :admin` option to our admin scope.
```
lancehalvorsen: Please add a comma before "which is not correct".

C_routing.md
```diff
@@ -532,27 +520,27 @@ end
...
   api_v1_user_path  GET     /api/v1/users/:id         HelloPhoenix.Api.V1.UserController.show/2
   api_v1_user_path  POST    /api/v1/users             HelloPhoenix.Api.V1.UserController.create/2
-  api_v1_user_path  PUT     /api/v1/users/:id         HelloPhoenix.Api.V1.UserController.update/2
-                    PATCH   /api/v1/users/:id         HelloPhoenix.Api.V1.UserController.update/2
+  api_v1_user_path  PATCH   /api/v1/users/:id         HelloPhoenix.Api.V1.UserController.update/2
+                    PUT     /api/v1/users/:id         HelloPhoenix.Api.V1.UserController.update/2
   api_v1_user_path  DELETE  /api/v1/users/:id         HelloPhoenix.Api.V1.UserController.delete/2
 ```
-Interestingly, we can re-define the same scope as long as we are careful not to duplicate routes. If we do duplicate a route, we'll get this familiar warning.
+Interestingly, we can use multiple scopes with the same path as long as we are careful not to duplicate routes (resources). If we do duplicate a route, we'll get this familiar warning.
```
lancehalvorsen: Please remove the "(resources)" part here.

C_routing.md
```diff
@@ -619,12 +592,18 @@ We have come quite a long way in this guide without talking about one of the fir
 
 Remember in the [Overview Guide](http://www.phoenixframework.org/docs/overview) when we described plugs as being stacked and executable in a pre-determined order, like a pipeline? Now we're going to take a closer look at how these plug stacks work in the router.
 
-Pipelines are simply plugs stacked up together in a specific order and given a name. They allow us to customize behaviors and transformations related to the handling of requests. Phoenix provides default pipelines for common tasks, but it allows us to customize them, and it also allows us to create new pipelines to meet our needs.
+Pipelines are simply plugs stacked up together in a specific order and given a name. They allow us to customize behaviors and transformations related to the handling of requests. Phoenix provides us with some default pipelines for a number of common tasks. In turn we can customize them as well as creating new pipelines to meet our needs.
```
lancehalvorsen: Please change "creating" to "create" for subject verb agreement purposes.

"In turn we can customize them as well as create new pipelines to meet our needs."

C_routing.md
```diff
@@ -619,12 +592,18 @@ We have come quite a long way in this guide without talking about one of the fir
...
 
-A newly generated Phoenix application defines two pipelines, `:browser`, and `:api`. We'll get to those in a minute, but first we need to talk about the plug stack in the Endpoint.
+A newly generated Phoenix application defines two pipelines called `:browser` and `:api`. We'll get to those in a minute, but first we need to talk about the plug stack in the Endpoint plugs.
 
 #####The Endpoint Plugs
 
+The endpoint is the boundary where all requests start, thereby exposing the application interface to the underlying web services. It has three responsibilities:
```
lancehalvorsen: I don't think we can just use text verbatim from the docs without any attribution. We either need to re-write this section, add attribution, or remove it.

C_routing.md
```diff
@@ -649,17 +628,17 @@ Older versions of Phoenix defined a third pipeline `:before`. Its purpose was to
 
 Phoenix defines two other pipelines by default, `:browser` and `:api`. The router will invoke these after it matches a route, assuming we have called `pipe_through/1` with them in the enclosing scope.
 
-As their names suggest, the `:browser` pipeline prepares for routes which render HTML for a browser. The `:api` pipeline prepares for routes which produce data for an api.
+As their names suggest, the `:browser` pipeline prepares for routes which render HTML for a browser. The `:api` pipeline prepares for routes which produce data for an api, for example JSON.
```
lancehalvorsen: I don't think that the ", for example JSON" actually adds anything here. Please remove it.

C_routing.md
```diff
@@ -649,17 +628,17 @@ Older versions of Phoenix defined a third pipeline `:before`. Its purpose was to
...
 
 The `:browser` pipeline has four plugs: `plug :accepts, ~w(html)` which defines the request format or formats which will be accepted, `:fetch_session`, which, naturally, fetches the session data and makes it available in the connection, `:fetch_flash` which retrieves any flash messages which may have been set, and  `:protect_from_forgery`, which protects form posts from cross site forgery.
 
-Currently, the `:api` pipeline only defines `plug :accepts, ~w(json)`.
+Currently our example so far, the `:api` pipeline only defines `plug :accepts, ~w(json)`.
```
lancehalvorsen: Actually, the idea is that currently Phoenix only defines the `plug :accepts, ~w(json)`, not that our example does. 

Please remove "our example so far".

C_routing.md
```diff
@@ -649,17 +628,17 @@ Older versions of Phoenix defined a third pipeline `:before`. Its purpose was to
...
 
-The router will invoke a pipeline on a route defined within a scope. If no scope is defined, the router will invoke the pipeline on all the routes in the router. If we call `pipe_through/1` within a nested scope, the router will invoke it on the inner scope only.
+The router invokes a pipeline on a route defined within a scope. If no scope is defined, the router will invoke the pipeline on all the routes in the router. If from within a nested scope we call `pipe_through/1`, the router will invoke it on the inner scope only.
```
lancehalvorsen: I'm fine with changing "will invoke" to "invokes", but I believe the word order of the next change is awkward. We can mix our ideas with the following wording:

"If we call `pipe_through/1` from within a nested scope,"

C_routing.md
```diff
@@ -649,17 +628,17 @@ Older versions of Phoenix defined a third pipeline `:before`. Its purpose was to
...
 
-Those are a lot of words bunched up together. Let's take a look at some examples to untangle their meaning.
+That's quite a few words bunched up together, so let's take a closer look at some examples to untangle their meaning and clarify things.
```
lancehalvorsen: Actually, please leave the original version here.

C_routing.md
```diff
@@ -769,10 +749,11 @@ defmodule HelloPhoenix.Router do
 end
 ```
 
-In general, the scoping rules for pipelines behave as you might expect. In this example, all routes will pipe through the `:browser` pipeline, because the `/` scope encloses all the routes. Only the `reviews` resources routes will pipe through the `:review_checks` pipeline, however, because we declare `pipe_through :review_checks` within the `/reviews` scope, where the `reviews` resources routes are.
+In general, the scoping rules for pipelines behave as you would expect. In this example, all routes will pipe through the `:browser` pipeline, because the `/` scope encloses all the routes. Only the `reviews` resources routes will additionally pipe through the `:review_checks` pipeline, because we declare `pipe_through :review_checks` within the `/reviews` scope, where the `reviews` resources routes are.
```
lancehalvorsen: I actually like the original version better here.

C_routing.md
```diff
@@ -816,7 +797,8 @@ defmodule HelloPhoenix.Router do
   ...
 end
 ```
-Next, we need to use the `channel/3` macro to match a topic to the channel which will handle its requests. If we have a channel module called `RoomChannel` and a topic called `"rooms:*"`, the code to do this is straightforward.
+
```
lancehalvorsen: I think it's very important to keep the original here for two reasons.
- We need to call out that `channel/3` is a macro, just like `get/3` or `resources/3` above.
- Our topic is not 'lobby' . The topic is 'rooms:*'.

C_routing.md
```diff
@@ -906,10 +896,10 @@ end
 
 Routing is a big topic, and we have covered a lot of ground here. The important points to take away from this guide are:
 - Routes which begin with an HTTP verb name expand to a single clause of the match function.
-- Routes which begin with 'resources' expand to 8 clauses of the match function.
+- Routes which begin with 'resources' expand to 8 clauses of the match function: GET, POST, PUT, PATCH and DELETE.
```
lancehalvorsen: We need to go with the original here for two reasons.
- The capitalized versions of GET, POST, etc. are the HTTP verbs, not clauses for our match function.
- If you count, there are only five verbs listed, and we mention eight clauses. :^)

lancehalvorsen:
@kgish thanks for such a large amount of work! Thanks also for re-working the PR to make allow it to be merged. All is well on that front. As you can see, I have left a number of comments.

C_routing.md
kgish: OK.

C_routing.md
kgish: Ok.

C_routing.md
kgish: Ok.

C_routing.md
kgish: Ok.

C_routing.md
kgish: Ok.

C_routing.md
kgish: Removed it.

C_routing.md
kgish: Removed.

C_routing.md
kgish: Ok.

C_routing.md
kgish: Ok.

C_routing.md
kgish: Ok.

C_routing.md
kgish: Ok, if you insist.

C_routing.md
kgish: Ok, must have missed that one.

C_routing.md
kgish: Ok.

kgish:
Updated according to your suggestions and pushed so you should be able to merge.

lancehalvorsen:
@kgish this is awesome, thanks! <3 <3 <3

kgish:
You're welcome, I'll see about proof reading the following chapters soon and doing my best to follow your guidelines.

PR #150, opened 2015-03-04
Remove obsolete text on setting `server: false`

shankardevy:
As of v.0.9, I don't see it's necessary to set server: false. Maybe I am wrong. Making just a suggestion here.


chrismccord:
This didn't change, and it still applies. See here https://github.com/phoenixframework/phoenix/blob/master/priv/template/config/prod.exs#L29-L40

PR #156, opened 2015-03-10
refurb for v0.10.0

lancehalvorsen:
- `mix phoenix.new` only takes one argument
- the task prompts us to install mix dependencies as well as brunch.io dependencies


lancehalvorsen:
@chrismccord funny, we were working on the same file at the same time. :) I'll push these changes to the site right away.

PR #162, opened 2015-03-10
Fix `use` line in new HelloView

avidal:
I submitted this as a suggestion against readme.io already, but @jeregrine on IRC suggested I do it here.

I got the new `use` line by looking at the generated views after getting a compilation error using the original way.


lancehalvorsen:
@avidal thanks for this! I'm reworking this whole guide at the moment and this change will be included. Look for an update later this evening. Thanks again!

PR #166, opened 2015-03-13
Fix nginx setup steps

mosic:
The original line symlinked whole `sites-available` directory, which results in nginx giving an error:

```
"/etc/nginx/sites-enabled/sites-available" failed (21: Is a directory)
```


lancehalvorsen:
Thanks @mosic!

PR #170, opened 2015-03-18
Use _url in guides

josevalim:


lancehalvorsen:
Thanks @josevalim! Good to see you here. :)

Issue #175, opened 2015-03-21
Channels section talks about removed subscribers function


ijt:
The first reference is

    Phoenix.PubSub.subscribers(HelloPhoenix.PubSub, "rooms:demo")

Some rewriting around it will be needed.

lancehalvorsen:
Thanks for reporting this, @ijt! I'm still catching up to the 0.10.0 changes, and this is now on the list.

If you feel like helping out, pull requests are always welcome. :)

lancehalvorsen:
PR #176 closes this.

ijt:
Got it. Thanks Lance.

lancehalvorsen:
:+1:

chrismccord:
I will have some edits around here soon. subscribers is an internal API that really no-one should touch other than pubsub adapter authors.

lancehalvorsen:
@chrismccord much appreciated in advance! :)

Btw, is there an exposed/approved way to check which pids are subscribed to a topic on a channel? I clearly missed it if there is.

PR #176, opened 2015-03-25
refurbish channel guide for v0.10.0

lancehalvorsen:
- fixes the immediately broken instructions
- still missing broadcasting to external topics (via controllers and channels), which will be handled in another PR


lancehalvorsen:
closes #175

PR #181, opened 2015-03-31
Update inconsistencies in E_views.md

pel-daniel:
Looks like something changed between versions of Phoenix and some outputs where not updated


lancehalvorsen:
@pel-daniel thanks again!

PR #184, opened 2015-04-07
Update Channel Guide to include js client / chat example

chrismccord:


lancehalvorsen:
Great, great thanks for tackling this! I'll get a chance to look at the changes tonight.

chrismccord:
Merging this now since 0.11.0 is out and we need up to date examples. We can tweak/grow this as is, but it's better to have update-to-date info than out :)

lancehalvorsen:
:) You absolutely have a point there. :)

PR #192, opened 2015-04-11
Re-write the Ecto Models Guide

lancehalvorsen:
Complete top-to-bottom re-write to reflect all the changes in ecto over the last couple of months.

I'm going to let this simmer here for a day or so before I push it to the site. Comments welcome!


lancehalvorsen:
@chrismccord you left a nice comment in here somewhere as well, which I no longer see, but thanks!

PR #200, opened 2015-04-15
Fix typos in G_channels.md

pel-daniel:


lancehalvorsen:
@pel-daniel thanks again for giving this a close reading!

PR #202, opened 2015-04-20
Improvements to ecto guide

elliotcm:
This PR contains a couple of typo fixes and some larger changes to provide more culturally accepting examples.

### Remove "clearly", "obviously", "essentially"
Using language like "clearly", "obviously", and
"basically" is a common habit and is not intended
to be negative, but it can be offputting when
found in guidance as the concepts described are
frequently neither clear nor obvious for newcomers.

It's also redundant, as if it were obvious we wouldn't
be putting it in the documentation.

### Switch validation example away from surname length
This change removes the split of "first name" and
"last name" with just a "name" field, and adds a
"bio" field to use as an example for length
validation.

There's two main reasons for this change:

1. The idea of a neat first name and last name
   is not as universal as we might think, and the
   benefit gained from having separate fields is
   small against the cost of confusion and annoyance.
2. There's no practical reason for putting a length
   limit on a name field, especially one shorter
   than the database field it's going into.  It's
   not a great experience for those of us with
   long surnames.

These sorts of cultural assumptions are not
implemented with malice in mind, just habit,
but they provide little benefit to us as programmers
or to our products, and often only serve to make
those from other cultures feel unwelcome.

We should wherever possible avoid spreading these
antipatterns to other developers.

There's some more discussion on name fields here:
https://www.gov.uk/service-manual/user-centred-design/resources/patterns/names.html

### DIFF
--- a/H_ecto-models.md
+++ b/H_ecto-models.md
@@ -189,15 +189,15 @@ defmodule HelloPhoenix.User do
   use HelloPhoenix.Web, :model
 
   schema "users" do
-    field :first_name, :string
-    field :last_name, :string
+    field :name, :string
     field :email, :string
+    field :bio, :string
     field :number_of_pets, :integer
 
     timestamps
   end
 
-  @required_fields ~w(first_name last_name email number_of_pets)
+  @required_fields ~w(name email bio number_of_pets)
   @optional_fields ~w()
 
   @doc """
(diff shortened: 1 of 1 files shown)


jeregrine:
:+1:

lancehalvorsen:
Thanks for your input and work on this, @elliotcm!

PR #216, opened 2015-05-04
Be Explicit About Dependencies

### LINKED ISSUE #210: Explicitly cover Phoenix dependencies
I am not sure if this content exists yet, but when we install Phoenix, maybe we should say explicitly the tools we will be using in the guides.

* node.js (for brunch.io used for assets)
* postgres
* erlang
* elixir
* [inotify-tools](https://github.com/rvoicilas/inotify-tools/wiki) if you are on linux (needed for live reload)

lancehalvorsen:
- add Dependencies section to the Overview Guide
- ask users to read it before they begin in Up And Running

This closes #210 .

### DIFF
--- a/0_overview.md
+++ b/0_overview.md
@@ -66,3 +66,31 @@ Cowboy is an HTTP server written in Erlang by Loïc Hoguin of [99s](http://ninen
 Cowboy has fantastic documentation. The [Guides](http://ninenines.eu/docs/en/cowboy/HEAD/guide/) are especially helpful. Learning more about Cowboy will surely help you to understand Phoenix more fully.
 
 Cowboy has its own section of links in the [Resources Guide](http://www.phoenixframework.org/docs/resources).
+
+
+### System Dependencies
+
+There are a number of dependencies external to Phoenix which we will encounter as we work our way through these guides. Elixir and Erlang are hard dependencies, meaning that we won't be able to work with Phoenix at all unless we have them installed. The rest may not prevent us from getting started if we don't have them, but their absence may lead to errors that prevent us from moving forward. Simply installing these dependencies may save us from confusion and frustration later on.
+
+Let's take a look at each of them now.
+
+- Elixir
+Phoenix is written in Elixir, and our application code will also be written in Elixir. In order to do any work with Phoenix, we need Elixir installed on our system. The Elixir site itself has great [installation instructions](http://elixir-lang.org/install.html).
+
+- Erlang
+Elixir source code compiles to Erlang byte code which runs on the Erlang Virtual Machine. That means we must have Erlang installed on our system - in addition to Elixir - in order to work with Phoenix. The Elixir site also has [Erlang installation instructions](http://elixir-lang.org/install.html#installing-erlang).
+
+- node.js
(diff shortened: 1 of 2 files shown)


0_overview.md
```diff
@@ -66,3 +66,63 @@ Cowboy is an HTTP server written in Erlang by Loïc Hoguin of [99s](http://ninen
 Cowboy has fantastic documentation. The [Guides](http://ninenines.eu/docs/en/cowboy/HEAD/guide/) are especially helpful. Learning more about Cowboy will surely help you to understand Phoenix more fully.
 
 Cowboy has its own section of links in the [Resources Guide](http://www.phoenixframework.org/docs/resources).
+
+
+### Dependencies
```
josevalim: What if we call this "System dependencies"? What I mean is, dependencies you need to install on your machine and that are not brought as part of Phoenix. We can start with the operating system, saying we support all of them. And then continue as is?

0_overview.md
```diff
@@ -66,3 +66,63 @@ Cowboy is an HTTP server written in Erlang by Loïc Hoguin of [99s](http://ninen
...
+
+Let's take a look at each of them now.
+
+- Elixir
+Phoenix is written in Elixir, and our application code will also be written in Elixir. In order to do any work with Phoenix, we need Elixir installed on our system. The Elixir site itself has great [installation instructions](http://elixir-lang.org/install.html).
+
+- Erlang
+Elixir source code compiles to Erlang byte code which runs on the Erlang Virtual Machine. That means we must have Erlang installed on our system - in addition to Elixir - in order to work with Phoenix. The Elixir site also has [Erlang installation instructions](http://elixir-lang.org/install.html#installing-erlang).
+
+- node.js
```
josevalim: I think we can be more straight-forward here. We are going into deep into bits like `--no-brunch` that don't make sense until we create the first app. Just make it clear when you'd need it and that you can easily skip it and we are good. Something like:

> Node is an optional dependency. If you want to work with assets (javascript, css, etc), by default Phoenix will use brunch.io for asset compilation. Brunch.io uses the node package manager (npm) to install its dependencies, and npm requires node.js.

> If you don't care about assets or you want to use another build tool, pass the `--no-brunch` when creating a new application and node won't be required at all.

Thoughts?

0_overview.md
```diff
@@ -66,3 +66,63 @@ Cowboy is an HTTP server written in Erlang by Loïc Hoguin of [99s](http://ninen
...
+
+Install your brunch dependencies by running inside your app:
+
+$ npm install
+```
+
+When installing node.js, it is important to install a recent stable version. Currently, that means a version in the 0.12.x range. Some users have reported problems with significantly older versions of node - versions in the 0.10.x range, for example.
+
+- PostgreSQL
+PostgreSQL is a relational database server. When we work with Ecto models in these guides, we will use PostgreSQL and the Postgrex adapter for it. In order to follow along with the examples, we will need to install PostgreSQL on our system. The PostgreSQL wiki has [installation guides](https://wiki.postgresql.org/wiki/Detailed_installation_guides) for a number of different systems.
```
josevalim: We can mention one can use mysql by passing `--database mysql` when creating a new phoenix application.

josevalim:
This looks great! :+1:

lancehalvorsen:
Thanks for these great comments @josevalim! I was definitely working too hard on the node.js part. :)

I've incorporated all of your comments, with some slight wording changes, even though github can't quite see the node part.

PR #217, opened 2015-05-06
Add info on running phoenix in detached mode

shankardevy:


K_deployment.md
```diff
@@ -8,6 +8,15 @@ Now that we've got our application prepared for production, getting it running i
 MIX_ENV=prod PORT=4001 iex -S mix phoenix.server
 ```
 
+To run the application in detached mode
```
lancehalvorsen: We should probably complete this sentence, and maybe give a little context. :)

How about something like this:
We can run our application detached from the iex console. This effectively daemonizes the process so it can run independently in the background. The command to do that looks like the following.

K_deployment.md
```diff
@@ -8,6 +8,15 @@ Now that we've got our application prepared for production, getting it running i
...
+
+```elixir
+MIX_ENV=prod PORT=4001 elixir --detached -S mix phoenix.server
+```
+
+Running the application in detached mode allows you to keep the application running even after you terminate your shell connection with the server.
```
lancehalvorsen: One convention we have maintained in the guides is to substitute "we" and "our" for "you" and "your". It gives the subtle feeling that we're all in this together.

Making those substitutions here would be great, and I apologize for not having a contribution guide up yet. That would have covered this.

lancehalvorsen:
@shankardevy thanks so much for this! I left a couple of comments to address, but this is a good and useful addition.

lancehalvorsen:
@shankardevy ping

shankardevy:
@lancehalvorsen sorry for the delay. had been busy over the weekends.

lancehalvorsen:
Thanks!

PR #218, opened 2015-05-06
Use pipe operator more often

optikfluffel:
I've just watched over [the Plug guide](http://www.phoenixframework.org/v0.12.0/docs/understanding-plug) and found it a little weird that sometimes the pipe operator is used and sometimes not. It should be more consistent that way, don't you think?

### DIFF
--- a/DD_Plug.md
+++ b/DD_Plug.md
@@ -12,7 +12,7 @@ In order to act as a plug, a function simply needs to accept a connection struct
 ```elixir
 def put_headers(conn, key_values) do
   Enum.reduce key_values, conn, fn {k, v}, conn ->
-    Plug.Conn.put_resp_header(conn, k, v)
+    conn |> Plug.Conn.put_resp_header(k, v)
   end
 end
 ```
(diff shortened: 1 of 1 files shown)


lancehalvorsen:
@optikfluffel thanks for this.

TL;DR, I don't think that the pipe operator is a substitute for specifying the first argument in a single function call.

I believe that the usage here is correct. The pipe operator is useful when multiple functions are chained together, each depending on the return value of the last. If there is only one function call, the pipe doesn't buy us much. Here are some examples from Plug Guide.

This makes sense to me because there are multiple functions chained together: 
`conn |> put_flash("You must be logged in") |> redirect(to: "/")`

This also makes sense to me, assuming there are no other functions in the transformation:
`render conn, :show, page: find_page(params["id"])`

This does not seem like the most effective usage to me,  assuming there are no other functions in the transformation:
`conn |> render :show, page: find_page(params["id"])`

Does this explanation seem reasonable?

optikfluffel:
Ok thanks, then it's just a matter of taste I guess, thanks for the feedback ^^

PR #219, opened 2015-05-06
updated dependencies formating

davidgrupp:
When viewing In github the dependencies listed were poorly formatted and it was unclear which paragraph belonged to which item especially for the second paragraphs for node.js, postgres, and inotify-tools

### DIFF
--- a/0_overview.md
+++ b/0_overview.md
@@ -74,49 +74,23 @@ There are a number of dependencies external to Phoenix which we will encounter a
 
 Let's take a look at each of them now.
 
-<<<<<<< HEAD
 - Elixir
-
   Phoenix is written in Elixir, and our application code will also be written in Elixir. In order to do any work with Phoenix, we need Elixir installed on our system. The Elixir site itself has great [installation instructions](http://elixir-lang.org/install.html).
 
 - Erlang
-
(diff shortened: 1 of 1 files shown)


lancehalvorsen:
@davidgrupp thanks very much for pointing this out! Very helpful. 

I think that there is a better solution to this, which I've just pushed to master in this commit.
https://github.com/phoenixframework/phoenix_guides/commit/8e551b8f4e85a9569915765710daf7a12c342c31#diff-a504fa40c6709180037861059b9d82fb

What do you think?

lancehalvorsen:
The benefit, to me, of the alternate solution is that we can scan the list of slightly bolded dependencies and get the big picture at a glance.

davidgrupp:
@lancehalvorsen Yea I think the bolded dependencies are better, however, I like the indenting of mine. I kept the bullet points instead of using bold because it reflected the look of how the file originally looked in other markup renders and I was trying to make my changes minimal.

lancehalvorsen:
@davidgrupp you know what? I'm a dolt sometimes.

I only looked at the diff of your pr and didn't view the page as rendered on github. Now that I have looked at it, I like your solution better. Could I trouble you to rebase and re-push?

My apologies for the extra work.

davidgrupp:
@lancehalvorsen should be good to go now

lancehalvorsen:
@davidgrupp since there are a couple of extra commits here (mea culpa), I think a new PR would be ideal.

Thank you once again for your work on this. I truly appreciate it!

PR #220, opened 2015-05-07
updated dependencies formating

davidgrupp:
When viewing In github the dependencies listed were poorly formatted and it was unclear which paragraph belonged to which item especially for the second paragraphs for node.js, postgres, and inotify-tools


lancehalvorsen:
@davidgrupp Thanks so much!

davidgrupp:
@lancehalvorsen No problem, thank you!

PR #223, opened 2015-05-09
Switching to MySQL

lancehalvorsen:
I re-wrote an extended suggestion from the site helping people who might want to use MySQL instead of PostgreSQL.


H_ecto-models.md
```diff
@@ -609,3 +611,79 @@ end
 ```
 
 At this point, we are completely configured and ready to go. We can go back to the top of this guide and follow along.
+
+#### Using MySQL
```
josevalim: It may be worth noting that you can pass `--database mysql` to `phoenix.new`. Or do we mention it elsewhere?

lancehalvorsen: That's a great point. I don't think we do mention that anywhere. Thanks!

PR #225, opened 2015-05-10
Update definitions for Router and PageView.

jastix:
Changes were taken from newly generated Phoenix 0.12 application.


lancehalvorsen:
@jastix thanks for catching all of these!

PR #230, opened 2015-05-11
Fix typo in CONTRIBUTING.md

pel-daniel:


lancehalvorsen:
Actually, I believe that "beginning" is the correct usage. It's one of those cases where a gerund-like form can act as an adjective. For example, we would say an "aging population" instead of an "aged population". This is really a shortened form for "a population which is aging" or "users who are beginning".

pel-daniel:
Sorry, I didn't know about it. Thanks for the clarification !!

lancehalvorsen:
NP, I had a great grammar teacher in eighth grade. :)

PR #237, opened 2015-05-12
Include minimum nodejs requirement

chrismccord:


lancehalvorsen:
Thanks @chrismccord! I just made this change on the site as well.

PR #239, opened 2015-05-12
Use code (for examples) from latest phoenix version

ugisozols:
Some minor changes have happened so updated the guide to match what we get from latest version.


lancehalvorsen:
Thanks so much for catching these changes @ugisozols!

PR #243, opened 2015-05-13
Fix all headings so they should render correctly on Phoenix website

Nicd:
Headings should have a space between the # and the content, or they won't
render correctly on the Phoenix framework website.

### DIFF
--- a/B_adding_pages.md
+++ b/B_adding_pages.md
@@ -66,7 +66,7 @@ lib
 ```
 Enough prep, let's get on with our first new Phoenix page!
 
-###A New Route
+### A New Route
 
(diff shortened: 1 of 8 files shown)


jeregrine:
My scan says this is good. @lancehalvorsen do you think there will be an issue with the ecto ones? (I don't think so I'm just not certain)

lancehalvorsen:
@Nicd thanks for this! You saved me from having to do it tonight. :)

@jeregrine I just tested the mix task `mix something` ones on the site - those were the ones I thought might not work. All was well.

jeregrine:
<3

PR #244, opened 2015-05-14
Note that phoenix.gen.html also generates test files.

jastix:
Task `mix phoenix.gen.html` also generates a model and a controller tests.


lancehalvorsen:
Thanks for this, @jastix!

PR #248, opened 2015-05-16
Initial testing channels guide

gjaldon:
Far from finished but just putting it up here so you could 'course-correct' if necessary @lancehalvorsen. :) 

By the way, how to put this 'testing channels' guide after the 'channels guide'? Would I have to rename the 'channels guide' file?


GG_testing_channels.md
```diff
@@ -0,0 +1,45 @@
+As developers we typically value tests since they help to 'future-proof' our applications by
+minimizing regression and provide updated documentation. Phoenix recognizes this and helps
+make it easier to write tests by providing conveniences for testing its different parts,
+including Channels.
+
+In the Channels guide it was mentioned that a "Channel" is a layered system with different
```
lancehalvorsen: What would you think about changing the first bit of this sentence to read something like this:

In the Channels Guide, we saw that . . .

GG_testing_channels.md
```diff
@@ -0,0 +1,45 @@
...
+functions may not be enough. We may want to verify that its different moving parts
+are working together as we expect. This integration testing would assure us that we
+correctly defined our channel route, the channel module, and its callbacks; and that
+the lower-level layers such as the PubSub and Transport are configured correctly and
+are working as intended.
+
+#### The Channel Test Helpers Module
+
+When you generate a new Phoenix application, a `test/support/channel_case.ex` file is
+also generated for you. This file houses the `MyApp.ChannelCase` module which we will
```
lancehalvorsen: Perhaps change the "you" to "us"?

lancehalvorsen:
First of all, sincere thanks for taking this on @gjaldon! Good progress. :+1: 

To answer your question above, I have created a "testing" directory at the root of this project with a placeholder file for channels. If you could move this work there, that would be perfect.

On a general note, what would you think of mimicking the flow of the Ecto Models Guide? By that I mean, use the new channel generator to get both a working channel and working tests right out of the box. That gives readers a sense of accomplishment early - working stuff! It also provides concrete examples to work from.

Thanks again!

gjaldon:
Thanks for the feedback, @lancehalvorsen! I like the idea of using the channel generator early so users get concrete examples. Will be incorporating that into the guide.

gjaldon:
@lancehalvorsen this is ready for review. Let me know what you think!

lancehalvorsen:
@gjaldon thanks so much for your work on this! 

Here's my plan, as it stands now. I want all three of the planned testing guides to read well together, almost as if they were a single entity. I have ideas about how I want the others to go, but they aren't concrete yet. I'm going to merge this now, get the others done, and then perhaps edit this to read well with the others.

Thanks again, this is a great help!

gjaldon:
:+1:

Issue #169, opened 2015-03-13
Rethinking the Phoenix landing page


paulcsmith:
As discussed in IRC, the current website aesthetic is fairly generic and focuses a lot on technologies and not "human" reasons to switch. Compare the [phoenix homepage](http://www.phoenixframework.org) to [Ruby on Rails](http://rubyonrails.org) or [meteor](https://www.meteor.com) which focus less on technologies and more on what the framework has to offer developers. "Build apps that are a delight to use, faster than you ever thought possible" (Meteor), "Web development that doesn’t hurt - optimized
for programmer happiness and sustainable productivity" (Rails), compared to "A highly connected web framework" (Phoenix)

Goals for the redesign:

* More powerful branding. The current website blends in and doesn't have much emotion. I went with the Phoenix orange to make a more powerful impression.
* Emphasize the reasons for using Phoenix, not the technologies.
* Make it clear how to get to the guides
* Make it clear where people can go to get excited and see how an app is built.

**My questions for everyone are:**

* What made you switch to Phoenix?
* What did you switch from?
* Why did you switch?
* What led you to try Phoenix?

_Note:_ when I say "switch" that doesn't mean you use it exclusively. It just means you used to use one thing for a given problem and now you use Phoenix in that scenario.

For me, it was the speed and syntax of Elixir and Phoenix. Tests suites ran fast, dependencies installed quickly and the framework itself was ridiculously fast. That wouldn't have been enough though. It also need to have a beautiful syntax and nice APIs. This is the only framework/language combo I've found that covers all those things. So that's what I emphasized in the mockup. I wanted to make it clear that you get blazing fast speed without sacrificing a happy dev environment. 

Let me know what you think! Feel free to leave comments. This is very much a WIP. The real version would have real numbers and maybe even a chat with benchmarks along with some example code. I strongly believe that the screencast should also show some benchmarks at the end to make it clear that you're writing beautiful code *and* getting speed at the same time.

![phoenix v1](https://cloud.githubusercontent.com/assets/22394/6643304/a8305a46-c981-11e4-9614-b9f658278bbf.png)

P.S. The content below the blog line is very much a WIP. I'm mostly focus on the header section right now.

lancehalvorsen:
:+1: for upping our marketing game. Lot's of great ideas/insights here.

More later . . .

kgish:
Hi there.

I think that the reasons mentioned cover pretty much how I feel, so in 
that regard I do not have that much to add really.

[...]

PR #257, opened 2015-05-18
Add subheaders

josevalim:
Add subheaders and update the text as described in #169. I had to change the subheader texts when comported to #169 to avoid taking too much space but I think we are good go to.

One thing we need to improve is the call to action for "building the new web". Today they link to the channel guide but it would be nice if we pointed to a working app. Maybe we could have the source with something running on heroku. We don't need to think about it in the PR but I can open up another issue once we figure out a direction.

screen shot 2015-05-18 at 15 38 15


chrismccord:
@josevalim is amazing! Looks great to me. Lets :ship:

paulcsmith:
I love the three column heading and the text. Very nicely done :+1: I think those are good to ship as-is.

I'm still not a fan of *just* the "Productive. Reliable. Fast." heading though. One problem with the original page is that people didn't know what it was. So "Productive. Reliable. Fast" didn't relate to anything. What is productive, reliable and fast? And without a header saying what Phoenix is the next paragraph also doesn't make sense. What is productive? What can I compare this to? Rails, socket.io, what is it a framework for? I think until Phoenix becomes bigger it's important to make it clear that it's a framework for building web applications. Then draw them in with the intro paragraph and explain clearly in the three columns below that. 

My proposal is "Create reliable and fast web applications, (with ease / productively \ ) **or** including that it's a web framework in the intro paragraph:

"Reliable. Fast. Productive"

"Phoenix is a framework designed for creating web applications. Most languages and frameworks..."

styling/Phoenix_files/_hero.scss
```diff
@@ -84,3 +85,25 @@ body, h1, h2, h3, h4, h5, h6 {
   background-image: url("../Phoenix_files/phoenix-logo-white.png") !important;
   background-size: 250px 34px;
 }
+
+#subheader-columns {
```
paulcsmith: This should probably be a class

styling/Phoenix_files/_hero.scss
```diff
@@ -84,3 +85,25 @@ body, h1, h2, h3, h4, h5, h6 {
...
+  margin: 50px 30px 0; }
+
+.subheader h4 {
+  color: #000;
```
paulcsmith: This should probably introduce a new variable like `$body-color` so we can use it in multiple places. What do you think? I think it's in variables.css

styling/Phoenix_files/_hero.scss
```diff
@@ -84,3 +85,25 @@ body, h1, h2, h3, h4, h5, h6 {
...
+  font-weight: $light-font;
+  font-size: 18px;
+  padding-bottom: 14px;
+  text-transform: uppercase; }
+
+.subheader p {
+  font-weight: $light-font;
+  color: #333;
```
paulcsmith: Maybe a `$body-color` variable?

styling/Phoenix_files/_hero.scss
josevalim: I copied the `#category-columns` which is an id... To me it makes no difference, so if you have a strong feeling, let me know!

styling/Phoenix_files/_hero.scss
```diff
@@ -84,3 +85,25 @@ body, h1, h2, h3, h4, h5, h6 {
...
+  font-size: 16px; }
+
+.subheader a {
```
paulcsmith: I think it's good to avoid nesting if we can. Less breakage when html changes. Since we have full control over this html maybe we could add a class to the link like `subheader-link` What do you think about that?

styling/Phoenix_files/_hero.scss
josevalim: Agreed. will do.

paulcsmith:
@josevalim Way nice job, and awesome work on the design. Looks great! Left just a couple comments on the scss :)

styling/Phoenix_files/_hero.scss
```diff
@@ -84,3 +85,25 @@ body, h1, h2, h3, h4, h5, h6 {
...
+
+#subheader-columns {
+  margin: 50px 30px 0; }
+
+.subheader h4 {
+  color: #000;
+  font-weight: $light-font;
+  font-size: 18px;
+  padding-bottom: 14px;
+  text-transform: uppercase; }
```
paulcsmith: In this file, we're putting the closing curlies on a new line. It gets scrunched down when compiled though.

styling/Phoenix_files/_hero.scss
josevalim: The problem is that it is not really the color of the body... For example, the auto generated stuff is slightly lighter. Maybe $subheader-title-color and $subheader-color? For now I don't expect to re-use it anywhere... even though, is it a good practice to make them variables?

styling/Phoenix_files/_hero.scss
josevalim: Ah, sorry. Will fix it.

josevalim:
> And without a header saying what Phoenix is the next paragraph also doesn't make sense. What is productive? What can I compare this to? Rails, socket.io, what is it a framework for? I think until Phoenix becomes bigger it's important to make it clear that it's a framework for building web applications.

I would really like to try something quick and memorable as per the previous discussion. However, your points are completely valid. Here is building on your last suggestion:

> Productive. Reliable. Fast.
>
> Most web frameworks and languages make you choose between speed and a productive environment.
> Phoenix and Elixir gives you both.
>
> [check out the guides]
> [build apis, html5 apps, and more]

Some options:

1. Instead of making the last line (Phoenix and Elixir gives you both), we could make "web frameworks" & "languages" and "Phoenix" & "Elixir" in bold only.

2. Another idea is to remove Elixir from the headline so we can focus more on Phoenix

Other than I think this whole schema gives a great introduction. It is memorable, followed by our main value proposition and what you can build with it. Another option is to replace the last two lines: make "build apis, html5 apps and more" as the main focus on the button with "check out the guides" coming under with less focus.

Thoughts?

josevalim:
screen shot 2015-05-18 at 16 24 48

josevalim:
The screenshot above implements some of the suggestions in the previous comment.

paulcsmith:
I really like that a lot :+1: I like the removal of Elixir as well. It's not important in the intro paragraph.

I like the button change as well. Maybe we can shorten "Lean how with out guides" to just "See the guides" or something like that. Just a little less wordy. Other than that this is a fantastic change. Memorable, and makes it clear what Phoenix is for :+1:

styling/Phoenix_files/_hero.scss
paulcsmith: Yeah that's what I always do. Makes it easy to see all the colors in use and makes it easy to see if we're using too many different colors. I like those names :+1:

styling/Phoenix_files/_hero.scss
paulcsmith: Hmm, if readme is doing it then maybe we should stick with their convention. I'm cool with keeping it as an id for now :)

styling/Phoenix.html
```diff
@@ -35,7 +35,23 @@
...
+--></div></div></div></div></header>
+
+
+<div class="container body-container"><div><div id="subheader-columns"><div class="row">
+<div class="col-xs-4 subheader"><h4>How is Phoenix different?</h4>
+<p>Phoenix is a framework for building HTML5 apps, API backends and distributed systems. Written in Elixir, you get a beautiful syntax, a productive tooling and fast runtime.
```
paulcsmith: Another idea, bold the most important parts for quick scanning `<strong>beautiful syntax, a productive tooling and fast runtime</strong>`

styling/Phoenix.html
```diff
@@ -35,7 +35,23 @@
...
+<a href="http://www.phoenixframework.org/v0.10.0/docs">Get started with Phoenix</a></p></div>
+<div class="col-xs-4 subheader"><h4>Building the new web</h4>
+<p>Channels provide real-time streaming directly within Phoenix, allowing us to build rich, interactive applications, be it in the browser, a native mobile app or in embedded devices.
```
paulcsmith: `<strong>real-time streaming directly within Phoenix</strong>`

Also what do we mean "directly within Phoenix." Maybe that needs clarification?

lancehalvorsen:
Are you serious? @josevalim can do it all. :) :) :)

This looks fabulous to me. I did notice in the screenshot that the magnifying glass icon for search and the dropdown triangle to select version numbers are not showing up. Is this expected?

Other than that, I say :ship: :ship: :ship: :shipit:

styling/Phoenix.html
```diff
@@ -35,7 +35,23 @@
...
+<a href="http://www.phoenixframework.org/v0.10.0/docs">Learn about channels</a></p></p></div>
+<div class="col-xs-4 subheader"><h4>Battle-proven technology</h4>
+<p>Phoenix leverages the Erlang VM ability to handle millions of connections concurrently while also enjoying the VM guarantees for building fault-tolerant systems.
```
paulcsmith: `<strong>handle millions of connections concurrently<strong>` `<strong>guarantees for building fault-tolerant systems</strong>` 

These are all suggestions for what I think is most important, but feel free to disregard :)

paulcsmith:
@lancehalvorsen Yeah they show up when it's actually deployed on readme :)

lancehalvorsen:
Got it, thanks.

josevalim:
Thanks for the review and feedback @paulcsmith. I will merge this and apply it to readme.io! :heart_eyes_cat:

paulcsmith:
You’re welcome and thanks for doing it. I’m stoked to see this live! :D

josevalim:
It is live! :heart:

PR #267, opened 2015-05-24
Fix 'transport' sp. and add 'by' in Routing / Channel Routes.

tbloncar:


lancehalvorsen:
@tbloncar thanks very much!

tbloncar:
:+1:

PR #270, opened 2015-05-26
Testing Intro

lancehalvorsen:
This is the initial version of the guide, replacing the FPO.


testing/A_intruduction.md
```diff
@@ -0,0 +1,518 @@
...
+│   ├── channel_case.ex
+│   ├── conn_case.ex
+│   └── model_case.ex
+├── test_helper.exs
+└── views
+    ├── error_view_test.exs
+    └── page_view_test.exs
+```
+
+The test cases we get for free include `test/controllers/page_controller.exs`,  `test/views/error_view.exs`, and `test/views/page_view.exs`. Nice.
```
pel-daniel: The names of the files are missing the _test.exs in the name

testing/A_intruduction.md
```diff
@@ -0,0 +1,518 @@
...
+
+Finished in 0.2 seconds (0.2s on load, 0.00s on tests)
+3 tests, 0 failures
+
+Randomized with seed 220535
+```
+
+And we can run a single test in a file by appending a colon and a line number to the filename.
+
+Let's say we only wanted to run the test for the way `HelloPhoenix.ErrorView` renders `500.html. The test begins on line 12 of the file, so this is how we would do it.
```
ugisozols: Is there a missing ` for 500.html?

testing/A_intruduction.md
```diff
@@ -0,0 +1,518 @@
...
+
+...
+
+Finished in 0.1 seconds (0.1s on load, 0.00s on tests)
+3 tests, 0 failures
+
+Randomized with seed 125659
+```
+
+> Note: ExUnit tells us exactly which tags it is including and excluding for each test run. If we look back to the previous section on running tests, we'll see that line numbers specified for individual tests are actually treated as tags.
```
ugisozols: Just curios - did you really intended this to be a blockquote?

ugisozols: Ah, I see the same format for other notes further in the guide so I'm guessing it's intentional :)

lancehalvorsen: Intentional, yes. :) I'll have to clean up all the "Note:" instances in other guides for consistency, but I think it's more clear to readers this way.

PR #271, opened 2015-05-26
Fix Muddled Production Config in Advanced Deployment

### LINKED ISSUE #268: Advanced Deployment Guide overrides port configuration
Unless I'm missing something obvious, the way the advanced deployment guide recommends to set a default port doesn't work.

The guide currently recommends this:
```
config :hello_phoenix, HelloPhoenix.Endpoint,
  http: [port: {:system, "PORT"}],
  url: [host: "example.com"],
  cache_static_manifest: "priv/static/manifest.json",
  server: true,
  http: [port: 4001]
```

The wording makes me believe that it intends to set a default value of port 4001 when `PORT` is not passed in as an environment variable. However this _always_ sets the port to 4001, even when I specify the port via environment variable.

My understanding of exrm is too limited to recommend a better approach to this, but as far as I can tell the rest of the port configuration in this guide (like in the upstart file) doesn't do anything

lancehalvorsen:
- set the port statically to 8888
- offer guidance for setting the port via environment variable
- add instructions to run `MIX_ENV=prod mix phoenix.digest` before creating release

closes #268

### DIFF
--- a/L_advanced_deployment.md
+++ b/L_advanced_deployment.md
@@ -28,7 +28,7 @@ To get started, we'll need to add `{:exrm, "~> 0.15.3"}` into the list of depend
 
 ```elixir
   defp deps do
-    [{:phoenix, "~> 0.13"},
+    [{:phoenix, "~> 0.13.1"},
      {:phoenix_ecto, "~> 0.4"},
      {:postgrex, ">= 0.0.0"},
      {:phoenix_html, "~> 1.0"},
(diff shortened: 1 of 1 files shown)


citizenparker:
Just a comment on this - are you 100% you can override the port via environment variable at runtime with this approach? I believe this will get hardcoded at release compile time with exrm. The only way I could see to do this with exrm would be to set it via conform, but that would add a fair amount of complexity to this guide.

lancehalvorsen:
It works locally for me. Here's what I did to test.
- remove the current release
- switch `config/prod.exs` to use the PORT env var to set port
- `export PORT=4001`
- build prod release
- run prod release in console 
- check on port 4001 (it works)
- shut down the release
- `export PORT=4007`
- restart the release
- check on port 4007
- and it worked for me :)

PR #280, opened 2015-06-01
Fixing locale example to set conn var

jtmoulia:
Tho, I'm not sure what the indentation policy is for 90 char lines.


lancehalvorsen:
@jtmoulia thanks! I took a look at the preview and the line didn't break. For now, I think that's fine.

PR #281, opened 2015-06-03
Add nginx upgrade headers to prevent cutting connection.

mgwidmann:
I had trouble getting nginx not to cut the connection and return a 400 to the client without the following `Upgrade` and `Connection` headers.

### DIFF
--- a/deployment/I_exrm_releases
+++ b/deployment/I_exrm_releases
@@ -441,6 +441,10 @@ These are the contents of our `/etc/nginx/sites-available/hello_phoenix` file.
 upstream hello_phoenix {
     server 127.0.0.1:8888;
 }
+map $http_upgrade $connection_upgrade {
+    default upgrade;
+    '' close;
+}
 server{
     listen 80;
     server_name .hostname.com;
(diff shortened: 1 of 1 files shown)


lancehalvorsen:
@mgwidmann thanks for this! Question: is this specific to websocket requests for channels or generic HTTP requests?

If it is specific to channels, we should probably put some text around this to tell readers what it's for, and that it may not be necessary if they aren't using channels at all.

mgwidmann:
Yeah, regular requests worked fine. For channels to work, if Nginx is going to act as a websocket proxy, you must add that to the config. See their docs here: http://nginx.com/blog/websocket-nginx/ (down at the bottom).

mgwidmann:
Otherwise you have to tell the front-end code (Javascript or whatever) to connect directly to Phoenix instead of hitting the same domain/port as regular requests.

lancehalvorsen:
Indeed. 

I think it would be great if you could just add a line or two of explanation to the changes in this PR letting folks know what the extra config is about. Really, this would be a slightly re-worded version of the two comments you left above. My sense is that it would help people understand their purpose.

Thank you!

mgwidmann:
@lancehalvorsen Hows that?

lancehalvorsen:
:tada: Thanks so much @mgwidmann! :tada:

PR #286, opened 2015-06-10
Section on how to use has_many and belongs_to for ecto models

foxnewsnetwork:
Wrote a little blurb on how to declare related data models for your models. Please let me know if this is even the convention we are suppose to follow for doing relationships (I hope so, because it's how I've been doing it). If I'm off the mark, please let me know and I will dutifully correct.

Oh yeah, I have a question: what's the convention for deleting dependent models?

Should we alias the Repo and do it in the model using callbacks like so:
```elixir
defmodule App.User do
  use App.Web, :model
  alias App.Repo
  after_delete :destroy_posts
  schema "users" do
    has_many :posts, App.Post
  end

  def destroy_posts(%{model: model}) do
    Repo.delete_all model.posts
  end
end
```

Have the callback somewhere that already gets the Repo?
```elixir
defmodule App.User do
  use App.Web, :model
  after_delete App.UserController, :destroy_posts
  schema "users" do
    has_many :posts, App.Post
  end
end
```

Straight up do in delete action on the controller?
```elixir
defmodule App.UserController do
  def delete(conn, %{"id" => id}) do
    user = Repo.get(User, id) |> Repo.preload [:posts]
    Repo.delete_all(user.posts)
    Repo.delete(user)
    render(conn, "show.json", user: user)
  end
end
```

Some other way? 
Will we be getting:
```elixir
has_many :posts, App.Post, dependent: :destroy
```
Is this real life?

### DIFF
--- a/H_ecto-models.md
+++ b/H_ecto-models.md
@@ -517,6 +517,77 @@ end
 
 That's the end of our walk-through of Ecto usage in our controller actions. There is quite a bit more that Ecto models can do. Please take a look at the [Ecto documentation](http://hexdocs.pm/ecto/) for the rest of the story.
 
+### Data Relationship and Dependencies
+
+Handling individual tables is great, but if you want to build a modern web application, you will need a way to relate your data to each other. If you're a Ruby on Rails refugee, you'll be glad to know Ecto provides a very familiar API for building relationships between models:
+
+- Schema.has_many/3 : declares one to many relationships, for example, if you were building Youtube, an user might have many uploaded videos.
+- Schema.belongs_to/3 : declares a one to one relationship between parent and children models. In the Youtube example, an uploaded video belongs to its user.
+- Schema.has_one/3 : declares a one to one relationship. Note that has_one is just like has_many except instead of returning a collection of model structs, you get just one model struct. Continuing with the Youtube example, while an user might have many uploaded videos, the user might only have one featured video.
+
+```elixir
+# models/user.ex
+defmodule HelloPhoenix.User do
+  # other stuff
+  schema "users" do
+    field :name, :string
+    field :email, :string
+    field :bio, :string
+    field :number_of_pets, :integer
+
+    has_many :videos, HelloPhoenix.Video
+    timestamps
+  end
+  # more stuff
+end
+
+# models/video.ex
+defmodule HelloPhoenix.Video do
+  # other stuff
+  schema "videos" do
+    field :name, :string
+    field :approved_at, Ecto.DateTime
+    field :description, :text
+    field :likes, :integer
+    field :views, :integer
+
+    # if you don't provide a foreign_key, Ecto
+    # guesses it as the atom plus _id
+    belongs_to :user, HelloPhoenix.User, foreign_key: :user_id 
+
+    # If your table's column for indictating when 
+    # a row was first inserted is named "created_at", 
+    # you need to let Ecto know. ActiveRecord 
+    # immigrants in particular will need to do this as a 
+    # signal of allegience to Ecto
+    timestamps inserted_at: :created_at
+  end
+  @required_fields ~w(name approved_at description user_id)
+  # other stuff
+end
+```
+Note that you don't declare the field user_id in the video model, but, as a required (or optional) field, you do declare the user_id field.
+
+You can use your newly declared relationships in your controllers as thus:
+
+```elixir
+controllers/user_controller.ex
+defmodule Apiv2.UserController do
+  # other stuff
+  def index(conn, _params) do
+    users = User |> Repo.all |> Repo.preload [:videos]
+    render conn, "index.json", users: users
+  end
+
+  def show(conn, %{"id" => id}) do
+    user = User |> Repo.get(id) |> Repo.preload [:videos]
+    render conn, "show.json", user: user
+  end
+  # other stuff
+end
+```
+Notice the Repo.preload/2 function. Because we declared a relationship in user, %User{} will now contain a videos property which starts out as an unloaded relationship. In order to properly display it in render, you'll need to tell Ecto to preload the field. Note that Repo.preload is smart enough to work on just one model or collection of them.
+
 ### Integrating Ecto into an Existing Application
 
 Adding Ecto to a pre-existing Phoenix application is easy. Once we include Ecto and Postgrex as dependencies, there are mix tasks to help us.
--- a/H_ecto-models.md
+++ b/H_ecto-models.md
@@ -571,7 +571,7 @@ Note that we don't declare the field `user_id` in the video model, but, as a req
 We can use our newly declared relationships in our controllers as thus:
 
 ```elixir
-controllers/user_controller.ex
+# controllers/user_controller.ex
 defmodule HelloPhoenix.UserController do
 . . .
   def index(conn, _params) do
--- a/H_ecto-models.md
+++ b/H_ecto-models.md
@@ -522,7 +522,7 @@ That's the end of our walk-through of Ecto usage in our controller actions. Ther
 Suppose we are building a very simple video-sharing web application, in addition to having users on our site, we might also want to have videos. We asked Phoenix to scaffold a `Video` model for us:
 
 ```console
-$ mix phoenix.gen.model Video videos name:string approved_at:datetime description:text likes:integer views:integer
+$ mix phoenix.gen.model Video videos name:string approved_at:datetime description:text likes:integer views:integer user:belongs_to
 * creating priv/repo/migrations/20150611051558_create_video.exs
 * creating web/models/video.ex
 * creating test/models/video_test.exs
@@ -530,13 +530,13 @@ $ mix phoenix.gen.model Video videos name:string approved_at:datetime descriptio
 $ mix ecto.migrate
 ```
 
-Handling individual tables is great, but if we want to build a modern web application, we will need a way to relate our data to each other. Those of us with experience using Ruby's ActiveRecord will be glad to see that Ecto provides a very familiar API for building relationships between models:
+Handling individual tables is great, but if we want to build a modern web application, we will need a way to relate our data to each other. Those of us with experience using Ruby's ActiveRecord will be glad to see that Ecto provides a very familiar API for building relationships. For example,
 
 `Schema.has_many/3` declares one to many relationships, for example, in our video sharing service, our user model might have many uploaded video models.
 
 `Schema.belongs_to/3` declares a one to one relationship between parent and children models. In our video site example, an uploaded video belongs to its user.
 
-`Schema.has_one/3` declares a one to one relationship. Note that has_one is just like has_many except instead of returning a collection of model structs, we get just one model struct. Continuing with the video-sharing example, while an user might have many uploaded videos, the user might only have one featured video.
+`Schema.has_one/3` declares a one to one relationship. Note that has_one is just like has_many except instead of returning a collection of model structs, we get just one model struct. Continuing with the video-sharing example, while a user might have many uploaded videos, the user might only have one featured video.
 
 Here's how we would declare a `has_many` relationship in `web/models/user.ex`:
 ```elixir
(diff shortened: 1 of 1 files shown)


lancehalvorsen:
@foxnewsnetwork thanks so much! This is a really good addition. I'll leave some comments inline.

As to your question about current conventions for destroying dependent relationship records, I would check in with the Ecto mailing list: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!forum/elixir-ecto

H_ecto-models.md
```diff
@@ -517,6 +517,77 @@ end
...
 That's the end of our walk-through of Ecto usage in our controller actions. There is quite a bit more that Ecto models can do. Please take a look at the [Ecto documentation](http://hexdocs.pm/ecto/) for the rest of the story.
 
+### Data Relationship and Dependencies
+
+Handling individual tables is great, but if we want to build a modern web application, we will need a way to relate our data to each other. If we were Ruby on Rails refugees, we'll be glad to know Ecto provides a very familiar API for building relationships between models:
+
+`Schema.has_many/3` declares one to many relationships, for example, if we were building Youtube, our user model might have many uploaded video models.
+`Schema.belongs_to/3` declares a one to one relationship between parent and children models. In the Youtube example, an uploaded video belongs to its user.
+`Schema.has_one/3` declares a one to one relationship. Note that has_one is just like has_many except instead of returning a collection of model structs, we get just one model struct. Continuing with the Youtube example, while an user might have many uploaded videos, the user might only have one featured video.
+
```
lancehalvorsen: What would you think about putting in a migration somewhere in here for the Video model? My thought is that if people have been following along from the top of the guide, they should be able to put any code we give them into their project and have it work. Without a migration and `mix ecto.migrate`, that won't be the case for HelloPhoenix.Video.

H_ecto-models.md
```diff
@@ -517,6 +517,77 @@ end
 
 That's the end of our walk-through of Ecto usage in our controller actions. There is quite a bit more that Ecto models can do. Please take a look at the [Ecto documentation](http://hexdocs.pm/ecto/) for the rest of the story.
 
+### Data Relationship and Dependencies
+
+Handling individual tables is great, but if we want to build a modern web application, we will need a way to relate our data to each other. If we were Ruby on Rails refugees, we'll be glad to know Ecto provides a very familiar API for building relationships between models:
```
lancehalvorsen: I think "refugees" is an emotionally charged word we should be careful with. 

What would you think about changing the first part of that sentence to this?
Those of us with experience using ActiveRecord will be glad to see that Ecto provides . . .

H_ecto-models.md
```diff
@@ -517,6 +517,77 @@ end
...
+
+`Schema.has_many/3` declares one to many relationships, for example, if we were building Youtube, our user model might have many uploaded video models.
```
lancehalvorsen: What would you think about using a more generic equivalent to Youtube? What about this?

. . . if we were building a video sharing service, our . . .

H_ecto-models.md
```diff
@@ -517,6 +517,77 @@ end
...
+`Schema.belongs_to/3` declares a one to one relationship between parent and children models. In the Youtube example, an uploaded video belongs to its user.
```
lancehalvorsen: Here as well, what about:

In our video sharing service example, . . .

H_ecto-models.md
```diff
@@ -517,6 +517,77 @@ end
...
+    # signal of allegience to Ecto
+    timestamps inserted_at: :created_at
+  end
+  @required_fields ~w(name approved_at description user_id)
+. . .
+end
+```
+Note that we don't declare the field `user_id` in the video model, but, as a required (or optional) field, we do declare the `user_id` field.
+
+We can use our newly declared relationships in our controllers as thus:
```
lancehalvorsen: What would you think about changing the "as thus" to "like this"? To my ear, this seems more natural.

H_ecto-models.md
```diff
@@ -517,6 +517,77 @@ end
...
+### Data Relationship and Dependencies
+
+Handling individual tables is great, but if we want to build a modern web application, we will need a way to relate our data to each other. If we were Ruby on Rails refugees, we'll be glad to know Ecto provides a very familiar API for building relationships between models:
+
+`Schema.has_many/3` declares one to many relationships, for example, if we were building Youtube, our user model might have many uploaded video models.
+`Schema.belongs_to/3` declares a one to one relationship between parent and children models. In the Youtube example, an uploaded video belongs to its user.
+`Schema.has_one/3` declares a one to one relationship. Note that has_one is just like has_many except instead of returning a collection of model structs, we get just one model struct. Continuing with the Youtube example, while an user might have many uploaded videos, the user might only have one featured video.
+
+```elixir
+# models/user.ex
```
lancehalvorsen: Thank you for putting in the filenames. :)  In the rest of the guides, we haven't used code comments for this, though. We have put the filename above the elixir code fence inside of backticks. We might say something like:

Here's how we would declare a `has_many` relationship in `web/models/user.ex`.

H_ecto-models.md
```diff
@@ -517,6 +517,77 @@ end
...
+    field :bio, :string
+    field :number_of_pets, :integer
+
+    has_many :videos, HelloPhoenix.Video
+    timestamps
+  end
+. . .
+end
+
+# models/video.ex
```
lancehalvorsen: Same comment as line 529. That would mean that we would break this code up and have some plain text in between the code blocks.

H_ecto-models.md
```diff
@@ -517,6 +517,77 @@ end
...
+defmodule HelloPhoenix.Video do
+. . .
+  schema "videos" do
+    field :name, :string
+    field :approved_at, Ecto.DateTime
+    field :description, :text
+    field :likes, :integer
+    field :views, :integer
+
+    # If we don't provide a foreign_key, Ecto
```
lancehalvorsen: Here, too, we might want to break the code blocks up change these large comments into plain text.

H_ecto-models.md
```diff
@@ -517,6 +517,77 @@ end
...
+    # guesses it as the atom plus _id
+    belongs_to :user, HelloPhoenix.User, foreign_key: :user_id 
+
+    # If our table's column for indictating when
```
lancehalvorsen: Same comment as line 554.

H_ecto-models.md
```diff
@@ -517,6 +517,77 @@ end
...
+  @required_fields ~w(name approved_at description user_id)
+. . .
+end
+```
+Note that we don't declare the field `user_id` in the video model, but, as a required (or optional) field, we do declare the `user_id` field.
+
+We can use our newly declared relationships in our controllers as thus:
+
+```elixir
+# controllers/user_controller.ex
```
lancehalvorsen: Same comment as line 529.

lancehalvorsen:
Again, thanks so much for this contribution!

foxnewsnetwork:
Wow, thanks for all the feedback. I will implement them tonight after I go home.

lancehalvorsen:
:+1:

foxnewsnetwork:
ok, implemented your suggestions, everything actually does read a lot better now lol, thanks! If this is good, should I rebase?

lancehalvorsen:
First of all - love the changes. Thank you!

I spent a little time actually working through this section as a reader would, and I have just a few tiny suggestions. I'll put those inline now.

When those are resolved I think this will be ready to rebase, and we'll be all set.

Love this addition! <3

H_ecto-models.md
```diff
@@ -517,6 +517,86 @@ end
 
 That's the end of our walk-through of Ecto usage in our controller actions. There is quite a bit more that Ecto models can do. Please take a look at the [Ecto documentation](http://hexdocs.pm/ecto/) for the rest of the story.
 
+### Data Relationship and Dependencies
+
+Suppose we are building a very simple video-sharing web application, in addition to having users on our site, we might also want to have videos. We asked Phoenix to scaffold a `Video` model for us:
+
+```console
+$ mix phoenix.gen.model Video videos name:string approved_at:datetime description:text likes:integer views:integer
```
lancehalvorsen: What would you think about adding `user:belongs_to` at the end of this `phoenix.gen.model` call? It's just a little added insurance that there won't be any frustrating errors for users later on. That would make the whole call:

`$ mix phoenix.gen.model Video videos name:string approved_at:datetime description:text likes:integer views:integer user:belongs_to`

H_ecto-models.md
```diff
@@ -517,6 +517,86 @@ end
...
+$ mix ecto.migrate
+```
+
+Handling individual tables is great, but if we want to build a modern web application, we will need a way to relate our data to each other. Those of us with experience using Ruby's ActiveRecord will be glad to see that Ecto provides a very familiar API for building relationships between models:
+
+`Schema.has_many/3` declares one to many relationships, for example, in our video sharing service, our user model might have many uploaded video models.
+
+`Schema.belongs_to/3` declares a one to one relationship between parent and children models. In our video site example, an uploaded video belongs to its user.
+
+`Schema.has_one/3` declares a one to one relationship. Note that has_one is just like has_many except instead of returning a collection of model structs, we get just one model struct. Continuing with the video-sharing example, while an user might have many uploaded videos, the user might only have one featured video.
```
lancehalvorsen: Please change ". . . while an user . . . " to ". . . while a user . . . "

H_ecto-models.md
```diff
@@ -517,6 +517,86 @@ end
...
+* creating priv/repo/migrations/20150611051558_create_video.exs
+* creating web/models/video.ex
+* creating test/models/video_test.exs
+
+$ mix ecto.migrate
+```
+
+Handling individual tables is great, but if we want to build a modern web application, we will need a way to relate our data to each other. Those of us with experience using Ruby's ActiveRecord will be glad to see that Ecto provides a very familiar API for building relationships between models:
+
+`Schema.has_many/3` declares one to many relationships, for example, in our video sharing service, our user model might have many uploaded video models.
```
lancehalvorsen: I think the first sentence could end with "relationships". That would make it:

" . . . relationships. For example, . . . "

What do you think?

H_ecto-models.md
```diff
@@ -517,6 +517,86 @@ end
...
+Note that we don't declare the field `user_id` in the video model, but, as a required (or optional) field, we do declare the `user_id` field.
+
+We can use our newly declared relationships in our `web/controllers/user_controller.ex` like this:
+
+```elixir
+defmodule HelloPhoenix.UserController do
+. . .
+  def index(conn, _params) do
+    users = User |> Repo.all |> Repo.preload [:videos]
+    render conn, "index.json", users: users
```
lancehalvorsen: Up until this point, we've been working with an HTML resource. Using "index.json" here leads to a 500 page. Could you change it to "index.html", please?

H_ecto-models.md
```diff
@@ -517,6 +517,86 @@ end
...
+  end
+
+  def show(conn, %{"id" => id}) do
+    user = User |> Repo.get(id) |> Repo.preload [:videos]
+    render conn, "show.json", user: user
```
lancehalvorsen: This is the same situation as line 588. Could you change it to "show.html", please?

H_ecto-models.md
foxnewsnetwork: Whoa! I didn't know you could even do that. Thanks, will be using that in the future.

H_ecto-models.md
foxnewsnetwork: lol that's a good point, from the context it's obvious it's relationships between models.

H_ecto-models.md
foxnewsnetwork: Good catch.

foxnewsnetwork:
Implemented your suggestions. Please review and let me know if everything is good, then I will rebase.

lancehalvorsen:
This looks great! Please rebase and we're all set. <3

foxnewsnetwork:
@lancehalvorsen rebased and pushed. btw thanks helping me through the contrib process.

lancehalvorsen:
@foxnewsnetwork this is great. Thanks again so much! <3

PR #288, opened 2015-06-15
Fix path in Assigning Layouts section.

arianvp:
It was pointing to the wrong file.


lancehalvorsen:
@arianvp thanks very much!

PR #290, opened 2015-06-17
Update link to Template Guide.

arianvp:
The link was pointing to a non-existing anchor. I've fixed that.


lancehalvorsen:
@arianvp thanks so much!

PR #303, opened 2015-06-29
Sync with 0.14 release

chrismccord:
0.14 release is imminent and I believe I addressed all relevant changes. @lancehalvorsen we did away with `plug :action` and thus `plug :render`, so I reworked the examples around this. You might want to double check the flow. Thanks!


chrismccord:
Looks like my emacs `writeroom-mode` plugin made the commit really noisy with newlines. Sorry about that.

lancehalvorsen:
Thanks for the heads-up! I'll check through this and merge when I'm home.

PR #304, opened 2015-06-30
added section on working with lifecycle callbacks

foxnewsnetwork:
So I spent way too much time trying to beat callbacks into working in some app I'm working on. I couldn't find a guide to do this, so I wrote up this section after I figured it out.

### DIFF
--- a/H_ecto_models.md
+++ b/H_ecto_models.md
@@ -597,6 +597,60 @@ end
 ```
 Because we declared a relationship in `HelloPhoenix.User`, `%User{}` will now contain a videos property which starts out as an unloaded relationship. In order to properly display it in `render`, we'll need to tell Ecto to preload the field. Note that `Repo.preload/2` is smart enough to work on just one model or collection of them.
 
+### Working with Callback Hooks
+
+Continuing with our video sharing app example, many web applications require moderation for user submitted content. Our Video model has an `approved_at` field to signify that a moderator has approved the video. The question is, what happens if the user edits the video? Ideally, we'd like to be able to "null out" a video's `approved_at` field when this happens. Fortunately, Ecto provides us with a handful of life-cycle callbacks to achieve this.
+
+We will instruct Ecto to use the `before_update` hook in `web/models/video.ex` like so:
+
+```elixir
+defmodule HelloPhoenix.Video do
+. . .
+  before_update :reset_approved_at
+  def reset_approved_at(changeset) do
+    changeset
+    |> Ecto.Changeset.put_change(:approved_at, nil)
+  end
+end
+```
+Now, when `Repo.update` is called for the `Video` model, the `approved_at` field will automatically be reset to nil. This should keep us safe from the ravages of an often hostile modern Internet!
+
+`Ecto.Model.Callbacks` actually ships with with a lot more than the `before_update` callback, in addition, it comes with:
+
+- `before_delete`: called before the adapter deletes our record from the database; if perhaps we have a foreign_key constraint from another table, we would use this hook to clean it up.
+- `after_delete`: called after the adapter deletes our record. If we wished to archive our record, we'd do it in this hook. 
+- `before_update`: called before the adapter updates our record. We used it above to reset the likes counter.
+- `after_update`: called after the adapter updates our record. If we wanted to update our user that the model has changed, now would do that here.
+- `before_insert`: called before the adapter first creates a new record. We would use it in cases when we wanted to infer default values to certain fields from the partial changeset we receive from the user.
+- `after_insert`: called after the adapter creates our record. We could use this to notify the user a new has been created under that user's account.
+- `after_load`: called after the model is loaded from the database. We might use this to load in other values based upon what we found in our record.
+
+Note: there is no `before_load` hook.
+
+In all cases, a hook expects to be passed an atom that is the name of the function it is to call if that function is defined in the current module:
+
+```elixir
+before_insert :name_of_function_to_call
+def name_of_function_to_call(changeset) do
+  # do stuff with the changeset
+  # and return the modified changeset
+end
+```
+If our callback function lives elsewhere, hooks can take the name of the module as the first argument:
+```elixir
+before_insert HelloPhoenix.AnotherPlace, :name_of_function_to_call
+```
+
+Our callback function must always take a changeset as its first argument, and return a changeset. In our callbacks, we can expect to often be calling `Ecto.Changeset` functions like:
+
+- `put_change(changeset, key, value)`
+- `get_field(changeset, key, defaultValue//nil)`
+- `add_error(changeset, key, error)`
+
+For the full reference, visit the docs at [Ecto Changeset documentation](http://hexdocs.pm/ecto/Ecto.Changeset.html).
+
+Ecto callbacks give can be very powerful, but we need to take care when using them. On one hand, they can add behaviors in a way which may not be apparent in the execution flow because callbacks can be defined far from the action in our model. On the other hand, it might seem tempting to put a lot of behavior in a single callback, which can make the model difficult to test and debug.
+
 ### Integrating Ecto into an Existing Application
 
 Adding Ecto to a pre-existing Phoenix application is easy. Once we include Ecto and Postgrex as dependencies, there are mix tasks to help us.


lancehalvorsen:
Thanks so much for this @foxnewsnetwork! I should get a chance to review it in the next day or so.

H_ecto_models.md
```diff
@@ -597,6 +597,73 @@ end
 ```
 Because we declared a relationship in `HelloPhoenix.User`, `%User{}` will now contain a videos property which starts out as an unloaded relationship. In order to properly display it in `render`, we'll need to tell Ecto to preload the field. Note that `Repo.preload/2` is smart enough to work on just one model or collection of them.
 
+### Working with Callback Hooks
+
+Continuing with our video sharing app example, suppose some of our content-uploading users are horrible trolls and one of their favorite pastimes is to initially upload a video of happy huskies playing with furry kittens, then, after gathering a large number of "likes", change the happy video to an extremely controversial and blatantly offensive one. Then, they would write an inflammatory article about how the users of our app "like" extremely offensive videos, and spread this article around on social media to damage our online reputation and maybe even destroy business.
```
lancehalvorsen: I appreciate the humor. :) If we want to match the tone of the rest of the guide, though, we should probably re-think these first two paragraphs. :)

Here's an idea. This is just an idea. What if we focused on the `views` column instead? We could ask readers to imagine an `event` model with `type`, `model`, and `model_id` columns. Whenever a view event happens, an after_insert callback fires to update this column.

We wouldn't need to generate a whole new model because we aren't showing how the code behaves in a live system. We could show the event model with the schema and the callback.

What do you think?

H_ecto_models.md
```diff
@@ -597,6 +597,73 @@ end
...
+  before_update :reset_likes_count
+  def reset_likes_count(changeset) do
+    changeset
+    |> Ecto.Changeset.put_change(:likes, 0)
+  end
+end
+```
+Now, when `Repo.update` is called for the `Video` model, the `likes` field will automatically be reset to zero. This should keep us safe from the ravages of an often hostile modern Internet!
+
+`Ecto.Model.Callbacks` actually ships with with a lot more than the `before_update` callback, in addition, it comes with:
```
lancehalvorsen: I think this whole section describing the callbacks is really effective.

H_ecto_models.md
```diff
@@ -597,6 +597,73 @@ end
...
+
+- `before_delete`: called before the adapter deletes our record from the database; if perhaps we have a foreign_key constraint from another table, we would use this hook to clean it up.
+- `after_delete`: called after the adapter deletes our record. If we wished to archive our record, we'd do it in this hook. 
+- `before_update`: called before the adapter updates our record. We used it above to reset the likes counter.
+- `after_update`: called after the adapter updates our record. If we wanted to update our user that his model has changed, now would do that here.
+- `before_insert`: called before the adapter first creates a new record. We would use it in cases when we wanted to infer default values to certain fields from the partial changeset we receive from the user.
+- `after_insert`: called after the adapter creates our record. We could use this to notify the user a new has been created under his account.
```
lancehalvorsen: What would you think about using a more gender-neutral term  for the possessive pronoun here? By that I mean use "their" instead of "his"?

Note to self: I should add this to the contributing guidelines.

H_ecto_models.md
```diff
@@ -597,6 +597,73 @@ end
...
+- `after_load`: called after the model is loaded from the database. We might use this to load in other values based upon what we found in our record.
+
+Note: there is no `before_load` hook.
+
+In all cases, a hook expects to be passed an atom that is the name of the function it is to call if that function is defined in the current module:
```
lancehalvorsen: I think from here to the Ecto Changeset link is really effective.

foxnewsnetwork:
Thanks for the review and suggestions! I will make the changes tonight.

Great catch on the gender neutral thing, I really don't want to risk Phoenix / Elixir going the way of the nodejs community and toxifying due to political reasons.

H_ecto_models.md
```diff
@@ -597,6 +597,73 @@ end
...
+
+Our callback function must always take a changeset as its first argument, and return a changeset. In our callbacks, we can expect to often be calling `Ecto.Changeset` functions like:
+
+- `put_change(changeset, key, value)`
+- `get_field(changeset, key, defaultValue//nil)`
+- `add_error(changeset, key, error)`
+
+For the full reference, visit the docs at [Ecto Changeset documentation](http://hexdocs.pm/ecto/Ecto.Changeset.html).
+
+Finally, a word to the wise, callbacks and hooks give us great power to manipulate our business logic, but with great power comes great responsibility... and misusing that power will lead to an extremely messy code-base that's difficult to debug / test / develop (and lead to the death of your uncle). We should avoid stacking tons of IO-related "side-effects" into our callbacks. So avoid stuff like the following:
```
lancehalvorsen: I think this is an important paragraph, so I want to make sure that we're as simple an clear as possible. 

I took a first pass at an edit, and I'm curious what you think:
Ecto callbacks give can be very powerful, but we need to take care when using them. On one hand, they can add behaviors in a way which may not be apparent in the execution flow because callbacks can be defined far from the action in our model. On the other hand, it might seem tempting to put a lot of behavior in a single callback, which can make the model difficult to test and debug.

H_ecto_models.md
```diff
@@ -597,6 +597,73 @@ end
...
+
+```elixir
+# AVOID DOING THIS!
```
lancehalvorsen: Maybe we remove this comment and add an extra line to the paragraph above?

lancehalvorsen:
Thanks so much once again @foxnewsnetwork. I think this is a really important addition, and I'm looking forward to getting it in this guide!

foxnewsnetwork:
As usual, thanks for taking time to review this stuff! 

> I appreciate the humor. :) If we want to match the tone of the rest of the guide, though, we should 
> probably re-think these first two paragraphs. :)
>
> Here's an idea. This is just an idea. What if we focused on the views column instead? We could ask
>readers to imagine an event model with type, model, and model_id columns. Whenever a view event 
>happens, an after_insert callback fires to update this column.

lol you're absolutely right, I completely agree with keeping the guide professional. 

I like your "event model" suggestion and so I went back and re-read the other parts of the guide to look for fields that I could reset as an example callback response to an user event. It turned out `Video` has a `approved_at` field, and, to me, that seems like the most natural candidate to be reset when an user updates a record. I re-wrote the use-case motivation accordingly, please let me know what you think.

>I think this whole section describing the callbacks is really effective.

thanks :)

>What would you think about using a more gender-neutral term for the possessive pronoun here? By >that I mean use "their" instead of "his"?

Yes, definitely great catch. In this day and age, I need to be careful of my language lol

>I took a first pass at an edit, and I'm curious what you think:
Ecto callbacks give can be very powerful, but we need to take care when using them. On one hand, they can add behaviors in a way which may not be apparent in the execution flow because callbacks can be defined far from the action in our model. On the other hand, it might seem tempting to put a lot of behavior in a single callback, which can make the model difficult to test and debug.

Clean, professional, and gets the point across, I like it.

Anyway, I made all the edits according to your suggestions, please take another gander when you have the chance. Thanks

H_ecto_models.md
lancehalvorsen: I think we still need to fix the "his" here.

H_ecto_models.md
```diff
@@ -597,6 +597,62 @@ end
 ```
 Because we declared a relationship in `HelloPhoenix.User`, `%User{}` will now contain a videos property which starts out as an unloaded relationship. In order to properly display it in `render`, we'll need to tell Ecto to preload the field. Note that `Repo.preload/2` is smart enough to work on just one model or collection of them.
 
+### Working with Callback Hooks
+
+Continuing with our video sharing app example, suppose our service takes off and we now have a healthy community of viewers, content-creators, and a small team of volunteer moderators who review new uploads and approve the good ones while keeping out the illegal ones. But certain malicious users have figured out a way around our system by first uploading a "safe" video, waiting for our moderators to approve it, then changing the video to something ilegal.
```
lancehalvorsen: I like the idea of using the approved_at field!

I think we can express this idea more clearly with fewer words. Here's a quick first pass at the first two paragraphs. What do you think?

Many web applications require moderation for user submitted content. Our `Video` model has an `approved_at` field to signify that a moderator has approved the video. The question is, what happens if the user edits the video? Ideally, we'd like to be able to "null out" a video's `approved_at` field when this happens. Fortunately, Ecto provides us with a handful of life-cycle callbacks to achieve this.

foxnewsnetwork:
>Many web applications require moderation for user submitted content. Our Video model has an approved_at field to signify that a moderator has approved the video. The question is, what happens if the user edits the video? Ideally, we'd like to be able to "null out" a video's approved_at field when this happens. Fortunately, Ecto provides us with a handful of life-cycle callbacks to achieve this.

That sounds really concise. I like it.

By the way, happy July 4th, fellow Ameriburger

lancehalvorsen:
@foxnewsnetwork happy Fourth to you as well. :)

This looks really good! Lines 623 and 625 still have instances of "his" which should become "their". If you wanted to fix those and squash the commits, we're absolutely good to go. Otherwise, please squash and I'll fix them after I merge this in.

Thanks so much for this! <3 <3 <3

foxnewsnetwork:
neutered, squashed, rebased, and pushed. Please let me know if there's anything else I can do. Thanks.

lancehalvorsen:
:tada: This is great. Thanks so much @foxnewsnetwork!

PR #309, opened 2015-07-03
Link to the Mix Tasks guide for information on the test database.

wsmoak:
A user new to the project may not have Postgres running or the postgres role configured.
Move the note up to the point at which the tests will fail in that case, and link to the relevant info.

### DIFF
--- a/testing/A_introduction.md
+++ b/testing/A_introduction.md
@@ -6,6 +6,8 @@ ExUnit refers to a test module as a "test case", and we will do the same.
 
 Let's see this in action.
 
+> Note: Before we proceed, we'll need to have PostgreSQL installed and running on our system. We'll also need to configure our repo with the correct login credentials. [The section on ecto.create in the Mix Tasks guide](http://www.phoenixframework.org/docs/mix-tasks#section--ecto-create-) has more information on this, and the [Ecto Models Guide](http://www.phoenixframework.org/docs/ecto-models) dives into the details on how it all works.
+
 In a freshly generated application, let's run `mix test` at the root of the project. (Please see the [Up and Running Guide](http://www.phoenixframework.org/docs/up-and-running) for instructions on generating a new application.)
 
 ```console
@@ -23,6 +25,8 @@ Finished in 0.2 seconds (0.2s on load, 0.00s on tests)
 We already have four tests!
 
 In fact, we already have a directory structure completely set up for testing, including a test helper and support files.
+
+> Note: We didn't need to create or migrate our test database because the test helper took care of all that for us.
 
 ```console
 test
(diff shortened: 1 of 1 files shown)


testing/A_introduction.md
```diff
@@ -22,6 +22,8 @@ Finished in 0.2 seconds (0.2s on load, 0.00s on tests)
 
 We already have four tests!
 
+> Note: We don't need to create or migrate our test database because the test helper takes care of all that for us. If your tests fail to run and the error is related to Postgres, see [the section on ecto.create in the Mix Tasks guide](http://www.phoenixframework.org/docs/mix-tasks#section--ecto-create-) for more information.
```
lancehalvorsen: @wsmoak, thanks so much for this! Very much appreciated. :)

Could I ask one small favor? Please change the "your" to "the". That would become: "If the tests fail . . . ". With that and a rebase to squash the commits, we'll be all set!

(As much as possible, we try to avoid the use of "you" and "your" in the guides in favor of "we" and "our". My hope is that it gives readers a sense that we're all in it together. :) ) 

Thank you!

wsmoak:
Sure, not a problem.  I also noticed that there's another note around line 444 that says "Before we proceed, we'll need to have PostgreSQL installed..." and links to the Ecto Models Guide.

IME one _must_ have Postgres running and configured before `mix test` will do anything at all in the sample app generated by `mix phoenix.new myappname`.  Is that correct?

I'll read through it again and propose a different change, possibly combining the notes.  

Thanks for the quick feedback!

lancehalvorsen:
Thanks! You've identified an interesting thing here. My thought was to re-use the work done in the Ecto Models Guide for testing because it would be familiar territory. The link back to Ecto Models was intended to say, "If you've got questions, go there and get things set up. We'll be here when you're done." It doesn't appear that that is completely successful. I'm curious what you think.

If somebody runs `mix phoenix.new myappname`, yes, their new app will be configured for PostgreSQL, and they will need to have PostgreSQL installed on their system to work through this material.

While that's true, the Models Guide points to instructions for setting up MySQL as an alternative. (SQLite and MSSQL are also possibilities, but they aren't documented in these guides yet. MongoDB will also be a possibility soon.)

Please do give it another read-through and let me know if you have any other ideas. It's great to have feedback like this from people working through the guides, btw. It's something I can't do for myself, and folks on the core team know to much to act as new users. :^)

wsmoak:
I took another pass at it.  No problem if you want to close this and do something else entirely.

The issue is that a new user reading the testing/introduction page is sent to Up and Running to create a sample app, but when they return, `mix test` is going to fail if they don't have Postgres running and configured.

(Up and Running mentions that Postgres needs to be _installed_ but that isn't enough to make it work.)

lancehalvorsen:
I think this is really great. Thanks so much for this contribution @wsmoak!

PR #311, opened 2015-07-03
Add nodejs-legacy instructions for Debian users

nettofarah:
I'm not sure if this is the best place to add this paragraph to, but this is a pretty annoying error I've seen a few people running into.

Someone could argue that this is more of a brunch/nodejs problem, but I think it is fair that we provide some troubleshooting help since there's a decision to go with brunch as the default asset management tool.

### DIFF
--- a/introduction/E_installation.md
+++ b/introduction/E_installation.md
@@ -45,6 +45,24 @@ We can get node.js from the [download page](https://nodejs.org/download/). When 
 
 Mac OS X users can also install node.js via [homebrew](http://brew.sh/).
 
+Debin/Ubuntu users might see an error that looks like this:
+```console
+sh: 1: node: not found
+npm WARN This failure might be due to the use of legacy binary "node"
+```
+This due to Debian having conflicting binaries for node: [discussion on stackoverflow](http://stackoverflow.com/questions/21168141/can-not-install-packages-using-node-package-manager-in-ubuntu)
+
+There are two options to fix this problem, either:
+- install nodejs-legacy:
+```console
+$ apt-get install nodejs-legacy
+```
+or
+- create a symlink
+```console
+$ ln -s /usr/bin/nodejs /usr/bin/node
+```
+
 ### PostgreSQL
 
 PostgreSQL is a relational database server. Phoenix configures applications to use it by default, but we can switch to MySQL by passing the `--database mysql` flag when creating a new application.


introduction/E_installation.md
```diff
@@ -45,6 +45,24 @@ We can get node.js from the [download page](https://nodejs.org/download/). When
 
 Mac OS X users can also install node.js via [homebrew](http://brew.sh/).
 
+#### Debian/Ubuntu users
```
lancehalvorsen: Since we didn't create a new header for Mac OS X users just above, I don't think we should create a header here.

introduction/E_installation.md
```diff
@@ -45,6 +45,24 @@ We can get node.js from the [download page](https://nodejs.org/download/). When
...
+You might see an error that says something like:
```
lancehalvorsen: We can take the removed header from above and re-format this sentence to remove the "You", like this:

Debian/Ubuntu users might see an error that looks like this:

introduction/E_installation.md
```diff
@@ -45,6 +45,24 @@ We can get node.js from the [download page](https://nodejs.org/download/). When
...
+```console
+sh: 1: node: not found
+npm WARN This failure might be due to the use of legacy binary "node"
+```
+This due to Debian having conflicting binaries for node: [discussion on stackoverflow](http://stackoverflow.com/questions/21168141/can-not-install-packages-using-node-package-manager-in-ubuntu)
+
+There are two options to this:
```
lancehalvorsen: We could clarify this sentence a little to look like this:

There are two ways to fix this problem, either:

introduction/E_installation.md
```diff
@@ -45,6 +45,24 @@ We can get node.js from the [download page](https://nodejs.org/download/). When
...
+- installing nodejs-legacy:
```
lancehalvorsen: If we do the edit from the previous line, then "installing" should become "install"

introduction/E_installation.md
```diff
@@ -45,6 +45,24 @@ We can get node.js from the [download page](https://nodejs.org/download/). When
...
+```console
+$ apt-get install nodejs-legacy
+``` 
+- creating a symlink
```
lancehalvorsen: Again, if we do the edit from two lines above, "creating" would become "create"

introduction/E_installation.md
```diff
@@ -45,6 +45,24 @@ We can get node.js from the [download page](https://nodejs.org/download/). When
...
+sh: 1: node: not found
+npm WARN This failure might be due to the use of legacy binary "node"
+```
+This due to Debian having conflicting binaries for node: [discussion on stackoverflow](http://stackoverflow.com/questions/21168141/can-not-install-packages-using-node-package-manager-in-ubuntu)
+
+There are two options to this:
+- installing nodejs-legacy:
+```console
+$ apt-get install nodejs-legacy
+```
```
lancehalvorsen: Since we have an "either" up above, we'll need an "or" here.

lancehalvorsen:
@nettofarah thanks so much for reporting this and submitting this PR. I made a few small comments, but I think this is really important to get into the guides.

Thanks again!

introduction/E_installation.md
nettofarah: Sounds good!

nettofarah:
All good now, @lancehalvorsen.
Thanks!

lancehalvorsen:
Thanks so much for this contribution, @nettofarah! <3 <3 <3

PR #318, opened 2015-07-11
Note that io.js is not known to work in place of node.js

wsmoak:
A Windows user had trouble with this. It wasn't clear that their installation of io.js would not be sufficient to get Phoenix to work, even though it says it is npm compatible.

(We did not try to debug why io.js wasn't working, the user simply installed node.js and went on.  It's possible it *might* work, if someone puts in the time to figure it out.)


lancehalvorsen:
@wsmoak thanks very much for this.  <3

PR #321, opened 2015-07-15
Update ecto models with banged insert/update/delete

dbackeus:
Reflect that phoenix now generates controller using the banged methods for CUD operations.

Don't know if it ought to be noted somewhere why they're used but I feel someone with better knowledge of Phoenix should write that (I'm a first timer just going through the docs).


lancehalvorsen:
Good catch @dbackeus! Thanks very much for taking the time to send a PR. Much appreciated. <3

PR #327, opened 2015-07-20
Clarify Elixir/Erlang Dependency Installation

### LINKED ISSUE #326: Confusing wording in installation guide
> These are either Elixir or Erlang projects which are part of Phoenix applications by default. We won't need to do anything special to install them besides `mix deps.get`. If we let mix install our dependencies as we create our new application, we won't even need to do that.

A new user thought this meant they were supposed to run `mix debs.get` as the next step after archive.install. Can we tweak the wording here to clarify?

lancehalvorsen:
I added an extra section to Up and Running which should clarify this further, but I left it as a second commit in case it might lead to more issues.

Thoughts, @chrismccord?

closes #326

### DIFF
--- a/A_up_and_running.md
+++ b/A_up_and_running.md
@@ -56,6 +56,35 @@ $ mix phoenix.server
 
 > Note: if this is the first time you are running this command, Phoenix may also ask to install Rebar. Go ahead with the installation as Rebar is used to build Erlang packages.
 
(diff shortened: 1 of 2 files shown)


chrismccord:
:+1:

lancehalvorsen:
Thanks for checking!

PR #334, opened 2015-07-31
Extend the deployment guide

josevalim:
* We talk about secrets
* We talk about assets
* We provide an overall script to be used in prod

This needs to wait for the next Phoenix release because I use the new `mix phoenix.gen.secret`.


lancehalvorsen:
Thank you so much @josevalim. Is there a heart big enough for this? :heart: will have to do.

Since the `mix phoenix.gen.secret` task is only in master, I'm assuming that after merging, this guide should not be pushed to the site until the next Phoenix release, yes?

I will rephrase the "you" and "your" bits to match the style of the rest of the guides, but there is no hurry.

josevalim:
> Since the mix phoenix.gen.secret task is only in master, I'm assuming that after merging, this guide should not be pushed to the site until the next Phoenix release, yes?

Exactly. I want to get a release out soon though.

lancehalvorsen:
:+1:

Issue #332, opened 2015-07-30
The Heroku deployment guide is broken since HOST is nil on Heroku


peter:
I'm not sure what the proper fix for this is but the deploy didn't work for me since System.get_env("HOST") is nil on Heroku. I worked around this by doing:

heroku config:set HOST=myapp.herokuapp.com

Is the host needed in the prod config?

lancehalvorsen:
Hi @peter, I'm curious, did you use the app name that Heroku generated for you, or did you give it a custom one?

PR #336, opened 2015-08-01
Heroku: Add note on setting HOST and PORT config vars

jerodsanto:
When following the Heroku directions, I couldn't get my app to boot without first setting the `HOST` and `PORT` env vars.

I added a note to this effect, but I'm not sure if there's a better/more dynamic way of retrieving these from the Heroku environment without having to set them manually.

### DIFF
--- a/deployment/E_heroku.md
+++ b/deployment/E_heroku.md
@@ -180,6 +180,12 @@ Setting config vars and restarting mysterious-meadow-6277... done, v3
 SECRET_KEY_BASE: my_secret_key_base
 ```
 
+You will also need to define the `HOST` and `PORT` configs at this time:
+
+```console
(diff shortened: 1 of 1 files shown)


lancehalvorsen:
Thanks for this @jerodsanto! Before I merge this, it would be great to have one piece of information: did you use the app name that Heroku generated for you or did you provide your own?

I ask because I want to be really accurate with what we tell readers they need to do. Several of us have used these guides as they currently stand to successfully deploy to Heroku, but my guess is that we all used the generated name.

Thanks in advance for helping with this information!

jerodsanto:
I specified an app name in my app:create call. That may be the difference. 

-Jerod
(sent on the go)

lancehalvorsen:
Thanks! <3 I'll tweak this a little after the merge to reflect that.

closes #332

PR #340, opened 2015-08-07
fixes typo 'are' -> 'our'

PositiveControl:
From:  The modules responsible for rendering *are* views, and we'll make a new one of those next.
To:  The modules responsible for rendering *our* views, and we'll make a new one of those next.


lancehalvorsen:
Hi @PositiveControl, thanks for taking the time to submit a PR! Unfortunately, this might not be the typo you think it is. :)

Here's another way of reading that sentence:
Views are the modules responsible for rendering, and we'll make one of those next.

The version above might be more clear, but it seems more passive to me. I think the original is less so.

I'm not crazy in love with either version. If you have a strong opinion, I'm all ears.

PositiveControl:
Hey @lancehalvorsen, I appreciate the thorough reply/explanation.  It's all making sense now :)  I'll close this PR and seek another another opportunity to contribute.

Thanks again!

PR #343, opened 2015-08-08
import socket from deps/phoenix/web/static/js/phoenix

PositiveControl:
Without `import {Socket} from "deps/phoenix/web/static/js/phoenix"` in app.js, the browser console will raise:
```
ReferenceError: Can't find variable Socket
```
While importing directly into app.js seems to be the simplest solution, you could also:
```
import socket from "./socket"
```
Though I believe this would require a bit more tinkering and code rewriting to get it to work (a tad over my head right now).

Issues I've ran into with the alternative method - importing socket.js into app.js:

1. line 57 `let channel = socket.channel("topic:subtopic", {})` errors because of an unmatched topic.  Simple fix, just change the topic to: `"rooms:lobby"`
2. Looks like it's expecting a token.  A little over my head for this early in the docs :)
```
[Error] WebSocket connection to 'ws://localhost:4000/socket/websocket?token=undefined' 
failed: WebSocket is closed before the connection is established.
	disconnect (app.js, line 600)
	connect (app.js, line 615)
	(anonymous function) (app.js, line 1149)
	initModule (app.js, line 64)
	require (app.js, line 74)
	global code (app.js, line 1242)
```

### DIFF
--- a/G_channels.md
+++ b/G_channels.md
@@ -104,6 +104,8 @@ For our chat app, we'll allow anyone to join the `"rooms:lobby"` topic, but any 
 With our channel in place, lets head over to `web/static/js/app.js` and get the client and server talking.
 
 ```javascript
+import {Socket} from "deps/phoenix/web/static/js/phoenix"
+
 let socket = new Socket("/socket")
 socket.connect()
 let chan = socket.channel("rooms:lobby", {})


PositiveControl:
Here's a working repo of the tutorial with comments on the second solution (relevant lines 22 - 34):

https://github.com/PositiveControl/phoenix_chat/blob/master/web/static/js/app.js#L22

lancehalvorsen:
Hi @PositiveControl, thanks so much for this PR!  I took a look at the Phoenix Chat Example repo which has been updated for v0.16.0, and I don't believe we need to reach into the deps directory for our import. Please see this line:
https://github.com/chrismccord/phoenix_chat_example/blob/master/web/static/js/app.js#L1

Could you verify, please? Thanks again for your PR!

chrismccord:
We do need to update the guide to latest 0.16 standards as in this PR. The chat example was bumped to 0.16, but still uses the "old" vendor'd phoenix.js approach.

lancehalvorsen:
Thanks @chrismccord! Verified it is.  :^)  :^)  :^)

PR #346, opened 2015-08-10
Add table name to references

bratsche:
Update the example mix command to include the table name after references, otherwise mix will generate the error:

```
** (Mix) Phoenix generators expect the table to be given to user:references.
For example:

    mix phoenix.gen.model Comment comments body:text post_id:references:posts
```


lancehalvorsen:
Good catch @bratsche, thank you! <3

PR #349, opened 2015-08-12
Talking about Brunch plugins, mention TypeScript instead of JSX

leikind:
Because Babel already supports JSX, and Phoenix includes Babel support out of the box, it is misleading to mention JSX as something you can plugin if you need it. It is already there.


lancehalvorsen:
Great! Thank you so much @leikind. <3

PR #358, opened 2015-08-25
First version of Sending Email

lancehalvorsen:


bonus_guides/I_sending_email.md
```diff
@@ -0,0 +1,300 @@
...
+   {:phoenix_html, "~> 2.0"},
+   {:phoenix_live_reload, "~> 0.6", only: :dev},
+   {:cowboy, "~> 1.0"},
+   {:mailgun, "~> 0.1.1"}]
+end
+```
+
+Next, we'll need to run `mix deps.get` to bring the mailgun package into our application.
+
+We'll also need to configure our mailgun domain and api key in `config/config.ex`. Note that the domain is the full url that mailgun understands as our domain. It could be the one that mailgun provided us by default when we created our account, or it could be a custom domain we configured in mailgun.
```
chrismccord: let's use separate dev.ex and prod.secrets.exs for the api domain and key

bonus_guides/I_sending_email.md
```diff
@@ -0,0 +1,300 @@
...
+  Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+  """
+end
+. . .
+```
+
+# FIXME if we want to tidy things up a little, we can create a function for html as well
+
+```elixir
+  defp welcome_html
```
chrismccord: welcome_email and welcome_text will need to accept some state as passed in arguments

chrismccord:
@lancehalvorsen This looks great so far! One thing we need to make sure to cover is url helpers will require using the Endpoint instead of a `@conn` because we are render without the context of the connection.

lancehalvorsen:
Whoa, something strange in gitland here. Standby.

lancehalvorsen:
Ok, this is the version I believed that I had pushed earlier.

:+1: to clarifying the need to use the Endpoint instead of `@conn` for the helpers. I'll be on that tomorrow.

lancehalvorsen:
@chrismccord fixes in. Please have a look. Thanks!

bonus_guides/I_sending_email.md
```diff
@@ -0,0 +1,333 @@
...
+We'll also need to configure our mailgun domain and api key in `config/config.ex`.
+
+The mailgun domain will be a full url, something like this `https://api.mailgun.net/v3/sandbox-our-domain.mailgun.org`. The key will be a long string - "key-another-long-string". For security reasons, it's important to not commit these values to a public source code repository, so we'll use environment variables set in our system to bring them into our configuration.
+
+Here's how that would look in `config/config.exs`
+
+```elixir
+config :hello_phoenix,
+       mailgun_domain: System.get_env("MAILGUN_DOMAIN")
+       mailgun_key: System.get_env("MAILGUN_API_KEY")
```
jeregrine: this is a little problematic since we don't prescribe dot_env or foreman for running locally. Not sure of a better way though. Maybe in the secrets?

Also we need to mention this stuff in the heroku deploy guide to add to the elixir_buildpack.conf

lancehalvorsen: Oh, you're right - put it in the secrets. That's one of the comments Chris left that got blown away when I re-pushed last night.

Will fix.

jeregrine: Nice :)

lancehalvorsen: I thought about this last night and my sense is that we need to recommend something like a parallel "secrets" system for dev - `dev.secret.exs`, import it into `dev.exs`, and put `dev.secret.exs` in `.gitignore`.

Of course, this is in addition to putting the prod mailgun credentials in `prod.secret.exs` as well so that it will work in production.

Without `dev.secret.exs`, I don't see how we can keep the credentials out of source control and also out of environment variables for the dev environment.

Alternately, in order to remove duplication, we could suggest a `config.secret.exs`, import that into `config.exs`, and put that in `.gitignore`. This would be for credentials which are the same in development and production, like mailgun.

My apologies if that's what people were suggesting and I didn't get it. :)

Thoughts?

lancehalvorsen:
Here's another attempt. I've hedged and provided two options with a quick explanation of the pros and cons of each. I'm open to other approaches if anybody has a better idea.

// @jeregrine @chrismccord

chrismccord:
> Here's another attempt. I've hedged and provided two options with a quick explanation of the pros and cons of each. I'm open to other approaches if anybody has a better idea.

:+1: :shipit: :heart: :purple_heart: :yellow_heart: :heart:

lancehalvorsen:
Thanks @chrismccord! I've pushed this guide to phoenixframework.org in the bonus section.

A big, red :heart: back to you for all the delightful emoji! :)

PR #359, opened 2015-08-25
Make links more prominent. Closes #331

### LINKED ISSUE #331: Links on the guides are not obvious
The current style of the links, at least in the body of the document, makes them really hard to notice. They are not visible at a quick glance and sometimes are hard to notice due to the different shades of gray(ish) found in the text.

screen shot 2015-07-29 at 5 37 36 pm

IMHO, a simple underline would do wonders. Or even a background, but I am no designer.

screen shot 2015-07-29 at 5 40 54 pm

paulcsmith:


paulcsmith:
I've updated Readme.io already and it appears to have worked just fine

lancehalvorsen:
Thanks so much @paulcsmith!

chrismccord:
Thanks Paul! Can we also get a font-size bump for the sidebar and body content? It seems a bit small

PR #360, opened 2015-08-27
add callbacks to example socket

montague:
A few colleagues came to me asking why their websocket didn't work. They followed the guide to the letter, but in the `web/channels/user_socket.ex` example file, we omit the required callbacks and `transport` setting. This PR simply includes those missing bits of code so that a new user can follow the channel guide and get a working websocket implementation up and running without having to look for the code represented by the `...`.

### DIFF
--- a/G_channels.md
+++ b/G_channels.md
@@ -73,7 +73,12 @@ defmodule HelloPhoenix.UserSocket do
   use Phoenix.Socket
 
   channel "rooms:*", HelloPhoenix.RoomChannel
-  ...
+
+  transport :websocket, Phoenix.Transports.WebSocket
+
+  def connect(_params, socket), do: {:ok, socket}
+
+  def id(_socket), do: nil
 end
 ```
 


gamache:
Speaking as one of the colleagues, :+1:

G_channels.md
```diff
@@ -73,7 +73,12 @@ defmodule HelloPhoenix.UserSocket do
   use Phoenix.Socket
 
   channel "rooms:*", HelloPhoenix.RoomChannel
-  ...
+
+  transport :websocket, Phoenix.Transports.WebSocket
+
+  def connect(_, socket), do: {:ok, socket}
```
lancehalvorsen: How would you feel about changing `def connect(_, socket)` to `def connect(_params, socket)` to match the generated `UserSocket` module? I think it lets readers know what the argument contains, and still has the benefit of avoiding compiler warnings.

lancehalvorsen:
Thanks very much for this @montague! I left one teeny comment. With that resolved and squashed commits, we'll be good to go. <3

montague:
All set! Thanks for the quick response :)

lancehalvorsen:
Great! Thanks so much for your contribution @montague! <3

lancehalvorsen:
This is now live on phoenixframework.org.

montague:
sweet!!

PR #365, opened 2015-09-02
Create A_custom_primary_key.md

iguberman:
My motivation for this document: it is important to be able to create a primary key other than id:integer in a variety of situations, especially if Phoenix is being adopted in existing projects or organizations.  I wanted to have all the steps in one place and easily accessible to Phoenix users -- especially new Phoenix users like myself.   Big thanks to Jose Valim and Slack users  @sg, @angeloashmore, @gjaldon for their help and feedback!


HH_custom_primary_key.md
```diff
@@ -0,0 +1,97 @@
...
+        user.name
+   end
+end
+```
+And below is another modification we need to be able to specify name instead of id in the URL
+(again, based on https://github.com/gjaldon/phoenix_params)
+
+**2. Modify web/controllers/user_controller.ex show, update, and delete methods**
+
+NOTE the use of `Repo.get_by` instead of `Repo.get` and `name: name` instead of `id`
```
josevalim: Actually, you don't need to to those changes. `Repo.get!` will always use the primary key. So as long as you change the primary key, you are fine. :)

josevalim:
Thanks @iguberman! I have added some tiny comments which should improve your code (and take less work :D).

HH_custom_primary_key.md
```diff
@@ -0,0 +1,97 @@
+###Example: Create a model with custom primary key `name:string` instead of autogenerated `id:integer`
```
lancehalvorsen: Please remove the header. Readme.io puts one in for us, and a double header in that layout doesn't really work.

HH_custom_primary_key.md
```diff
@@ -0,0 +1,97 @@
...
+
+IMPORTANT NOTE: this example was chosen for simplicity.  It is not advisable to use something that is likely to change
```
lancehalvorsen: Please make text soft-wrap so that it flows well in readme.io. They will insert their own line breaks, and the text will become really choppy otherwise.

HH_custom_primary_key.md
```diff
@@ -0,0 +1,97 @@
+###Example: Create a model with custom primary key `name:string` instead of autogenerated `id:integer`
+
```
lancehalvorsen: What would you think about adding some intro text here? I think it would be good to give readers an idea of why we would choose to do this. What problem are we solving? Is this primarily for legacy systems or new?

It might be good to mention that while Ecto makes this possible, it's not necessarily the natural path Ecto wants us to take.

lancehalvorsen:
Thanks very much for this @iguberman! I think this is valuable information for the guides, and I want to fit it in. I'm just wondering where the best place is. If the Ecto project had its own guides, I think that would be best. I'd like to keep the main section for guides on core Phoenix elements. Ecto models slipped in because many, many people will need them.

How about this. Would you mind moving this file to `bonus_guides/EE_custom_primary_keys.md`, please?

I've left a few comments, and I may add just a few more - still thinking. :)  But again, thanks so much!

josevalim:
@lancehalvorsen we don't plan to have Ecto guides. So I would love if we could have an Ecto section here if you believe they fit. In particular, if we have an Ecto section, I would like to make this blog post into a guide: http://blog.plataformatec.com.br/2015/08/working-with-ecto-associations-and-embeds/

HH_custom_primary_key.md
iguberman: Perfect -- this makes necessary changes really trivial!

chrismccord:
> In particular, if we have an Ecto section, I would like to make this blog post into a guide: http://blog.plataformatec.com.br/2015/08/working-with-ecto-associations-and-embeds/

:+1: on Ecto section and :+1: on making that blog post into a guide.

lancehalvorsen:
Interesting. :)

I think that having Ecto guides here in their own section would be very convenient for folks, and it would solve the problem of where to put content like this. I'm fine with that idea.

Just to play devil's advocate, do you think that there are any negatives related to putting the guides for both projects in the same place? I mean that people might conflate the two projects as one?

chrismccord:
> Just to play devil's advocate, do you think that there are any negatives related to putting the guides for both projects in the same place? I mean that people might conflate the two projects as one?

This is definitely a problem we have today (the conflating), but I don't know if there's a good way around it. Given our generators are tightly coupled to Ecto, and we push Ecto by default, I think having a holistic approach to the guides makes sense, especially with no detailed Ecto material outside of the docs and blog posts.

josevalim:
> This is definitely a problem we have today (the conflating), but I don't know if there's a good way around it. 

I am actually not even sure if it is a problem. If you are starting with Phoenix, do we even care if you consider Phoenix and Ecto to be the same or different things? As long as the message is clear in the long term, I am fine with it.

iguberman:
@josevalim @lancehalvorsen I made all the changes you mentioned in your comments above (apologies for so many incremental commits :) )  Glad you find this guide useful and thanks so much for such prompt and great feedback!  :+1:

lancehalvorsen:
Hi @iguberman, thanks! Could I trouble you to squash those commits, please?

I may also massage some formatting/wording to match the rest of the guides. Please don't take offense. :^)

Much appreciated! <3

iguberman:
@lancehalvorsen Squashed all my commits for this PR.   Feel free to make changes to this document as you see fit!  Thank you very much! :)

lancehalvorsen:
Hi @iguberman, I think there's something amiss here. I'm seeing 21 commits at the moment. :) This might have something to do with the fact that you worked on this as two different users, but that's a guess.

Since this is a single, new file, maybe the thing to do is cut a fresh new branch off of master and manually add the new file in? Then create new pull request and I'll just close this one. What do you think?

iguberman:
Yeah, I know.  Sorry about that.   I was trying to squash all my commits
into one and something went wrong.   I'll create a new PR.

iguberman:
Replaced with #367

PR #368, opened 2015-09-04
Update D_controllers.md

larrylv:
The context is about using `_format` to select appropriate template, so we should use `:index` instead of `index.text`

### DIFF
--- a/D_controllers.md
+++ b/D_controllers.md
@@ -386,7 +386,7 @@ Of course, we can pass data into our template as well. Let's change our action t
 
 ```elixir
 def index(conn, params) do
-  render conn, "index.text", message: params["message"]
+  render conn, :index, message: params["message"]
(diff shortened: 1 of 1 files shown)


lancehalvorsen:
Hi @larrylv, thanks for submitting this PR! Both the atom form `:index` and the string form `"index.html"` work in this context. We do have a note just below this that mentions the atom form.

If we make this change, we'll have to make many more changes to this guide for consistency. The original decision was to go with explicit formats (the string form), and mention that the atom form also works. I personally still think that's a good decision.

Having said that, here is your chance to convince me otherwise! If you have a really good reason for this change, I'm all ears. :)

larrylv:
@lancehalvorsen Ha, I think you convince me. Let me close this pr. Thanks for you answer!

PR #369, opened 2015-09-05
C_routing.md edits

garyf:
includes elimination of a typo in the "multiple socket handlers" code example

### DIFF
--- a/C_routing.md
+++ b/C_routing.md
@@ -209,7 +209,7 @@ iex> user_path(Endpoint, :delete, 17)
 "/users/17"
 ```
 
-What about paths with query strings? Phoenix has you covered. By adding an optional fourth argument of key value pairs, the path helpers will return those pairs in the query string.
+What about paths with query strings? By adding an optional fourth argument of key value pairs, the path helpers will return those pairs in the query string.
 
 ```elixir
 iex> user_path(Endpoint, :show, 17, admin: true, active: false)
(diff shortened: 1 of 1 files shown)


C_routing.md
```diff
@@ -276,8 +276,6 @@ The paths to the user facing reviews would look like a standard resource.
 /reviews
 /reviews/1234
 /reviews/1234/edit
-
-and so on
```
lancehalvorsen: This should probably be the fake ellipsis that we use throughout the rest of the guides.

`. . .`

C_routing.md
```diff
@@ -286,8 +284,6 @@ The admin review paths could be prefixed with `/admin`.
 /admin/reviews
 /admin/reviews/1234
 /admin/reviews/1234/edit
-
```
lancehalvorsen: This should probably be the fake ellipsis as well.

C_routing.md
```diff
@@ -749,11 +745,11 @@ defmodule HelloPhoenix.Router do
 end
 ```
 
-In general, the scoping rules for pipelines behave as you might expect. In this example, all routes will pipe through the `:browser` pipeline, because the `/` scope encloses all the routes. Only the `reviews` resources routes will pipe through the `:review_checks` pipeline, however, because we declare `pipe_through :review_checks` within the `/reviews` scope, where the `reviews` resources routes are.
+In general, the scoping rules for pipelines behave as you might expect. In this example, all routes will pipe through the `:browser` pipeline, because the `/` scope encloses all the routes. However, only the `reviews` resources routes will pipe through the `:review_checks` pipeline, because we declare `pipe_through :review_checks` within the `/reviews` scope, where the `reviews` resources routes are located.
```
lancehalvorsen: I think we can ditch "However" here completely.

Only the `reviews` . . .

lancehalvorsen:
Thanks for your contribution @garyf! I left a few, small comments. If we can clear those up and squash the commits, we'll be good to go.

Btw, Routing was one of the first of all the guides. Since routing has been stable for quite a while, I haven't looked at most of this for a long, long time. It's really interesting to see how the voice and style have changed compared to the more recent guides. Thanks for holding up a mirror. :)

lancehalvorsen:
Hi @garyf, just checking in. Looks like we just have two tiny things left. Both are removing the "and so on" string and replacing it with three dots ". . ." With that and squashed commits we'll be all set!

garyf:
My apologies for overlooking your suggestion to squash the commits; da925b8 should do it

lancehalvorsen:
@garyf no problem. Thanks for rebasing! Once again, I appreciate you shining a light on these old parts of the guides. <3

PR #370, opened 2015-09-06
Update Ecto models guide

ugisozols:
~~I only followed the guide till "Data Relationship and Dependencies" so maybe that part also has something that's a bit out of date. Will try to look at it soon but in the meantime this is ready~~ :)

Besides some text changes/dup word removals I also updated console/code output as they appear in latest Phoenix (1.0.2).

### DIFF
--- a/H_ecto_models.md
+++ b/H_ecto_models.md
@@ -4,7 +4,7 @@ Most web applications today need some form of data storage. In the Elixir ecosys
 * MySQL
 * MSSQL
 * SQLite3
-* (coming soon) MongoDB
+* MongoDB
 
 Newly generated Phoenix applications integrate both Ecto and the PostgreSQL adapter by default.
 
@@ -28,13 +28,13 @@ $ mix phoenix.gen.html User users name:string email:string bio:string number_of_
 * creating web/views/user_view.ex
 * creating test/controllers/user_controller_test.exs
 
-Add the resource to the proper scope in web/router.ex:
+Add the resource to your browser scope in web/router.ex:
 
-resources "/users", UserController
+    resources "/users", UserController
 
 and then update your repository by running migrations:
 
-$ mix ecto.migrate
+    $ mix ecto.migrate
 ```
 
 Notice that we get a lot for free with this task - a migration, a controller, a controller test, a model, a model test, a view, and a number of templates.
@@ -215,7 +215,7 @@ defmodule HelloPhoenix.User do
   @doc """
   Creates a changeset based on the `model` and `params`.
 
-  If `params` are nil, an invalid changeset is returned
+  If no params are provided, an invalid changeset is returned
   with no validation performed.
   """
   def changeset(model, params \\ :empty) do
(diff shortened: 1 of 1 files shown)


H_ecto_models.md
```diff
@@ -561,20 +567,17 @@ defmodule HelloPhoenix.User do
 end
 ```
 
-And here is how we would declare the complementary `belongs_to` relationship in `web/models/video.ex`:
+Since we used generator to scaffold our `Video` model and specified user association with `user_id:references:users` `belongs_to` relationship will already be defined for us in `web/models/video.ex`:
```
ugisozols: When following the guide I was trying to add `belongs_to` to `web/models/video.ex` myself and didn't even notice that `belongs_to` was already there (thanks to generator). That's the reason I changed the text. @lancehalvorsen any comments/suggestions?

H_ecto_models.md
```diff
@@ -621,11 +624,11 @@ end
...
 Now, when `Repo.update` is called for the `Video` model, the `approved_at` field will automatically be reset to nil. This should keep us safe from the ravages of an often hostile modern Internet!
 
-`Ecto.Model.Callbacks` actually ships with with a lot more than the `before_update` callback, in addition, it comes with:
+`Ecto.Model.Callbacks` actually ships with a lot more than the `before_update` callback, in addition, it comes with:
 
 - `before_delete`: called before the adapter deletes our record from the database; if perhaps we have a foreign_key constraint from another table, we would use this hook to clean it up.
 - `after_delete`: called after the adapter deletes our record. If we wished to archive our record, we'd do it in this hook.
-- `before_update`: called before the adapter updates our record. We used it above to reset the likes counter.
+- `before_update`: called before the adapter updates our record. We used it above to "null out" a video's `approved_at` field.
 - `after_update`: called after the adapter updates our record. If we wanted to update our user that the model has changed, now would do that here.
```
ugisozols: I'm not sure what "If we wanted to update our user that the model has changed, now would do that here." is trying to say. Maybe something like "If we wanted to notify user that the model has changed, we would do that here."?

H_ecto_models.md
```diff
@@ -621,11 +624,11 @@ end
...
 - `before_insert`: called before the adapter first creates a new record. We would use it in cases when we wanted to infer default values to certain fields from the partial changeset we receive from the user.
 - `after_insert`: called after the adapter creates our record. We could use this to notify the user a new has been created under that user's account.
```
ugisozols: Maybe "We could use this to notify the user a new **video** has been created under that user's account."?

H_ecto_models.md
lancehalvorsen: I think this is great. The only thing I would change is to add some definite articles:

Since we used the generator . . . and specified the user association with `user_id:references:users`, the `belongs_to` . . .

H_ecto_models.md
```diff
@@ -582,7 +585,7 @@ defmodule HelloPhoenix.Video do
 . . .
 end
 ```
-Note that we don't declare the field `user_id` in the video model, but, as a required (or optional) field, we do declare the `user_id` field.
+Note that we don't declare the field `user_id` in the video model schema. We add it to required (or optional) field list instead.
```
lancehalvorsen: I think we could use a definite article here as well:

We add it to the required . . .

H_ecto_models.md
lancehalvorsen: I think it's trying to say that I didn't proofread enough. :)

I think your re-statement is perfect. <3

H_ecto_models.md
lancehalvorsen: :+1:

lancehalvorsen:
Thank you so much once again @ugisozols! I just left a few very small comments. With those addressed, we'll be good to go.

ugisozols:
Addressed comments and squashed commits. Ready to :ship: :)

lancehalvorsen:
Awesome! Thanks so much once again @ugisozols! Keeping the text up to date and finding awkward wording are both great services which I really, really appreciate. <3 <3 <3

ugisozols:
:purple_heart:

PR #375, opened 2015-09-15
issue 1214

iguberman:
Added the info from @josevalim response about how to get around this issue.

### DIFF
--- a/bonus_guides/EE_custom_primary_key.md
+++ b/bonus_guides/EE_custom_primary_key.md
@@ -61,14 +61,14 @@ defmodule HelloPhoenix.Repo.Migrations.CreatePlayer do
 end
 ```
 
-Let's move on to `web/models/player.ex` next. We'll need to add a module attribute `@primary_key {:name, :string, []}` describing our primary key as a string. We'll also need to remove the `field :name, :string` line because this is our new primary key. If this seems unusual, recall that the schema doesn't list the `id` field in models where `id` is the primary key.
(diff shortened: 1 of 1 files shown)


bonus_guides/EE_custom_primary_key.md
```diff
@@ -61,14 +61,14 @@ defmodule HelloPhoenix.Repo.Migrations.CreatePlayer do
 end
 ```
 
-Let's move on to `web/models/player.ex` next. We'll need to add a module attribute `@primary_key {:name, :string, []}` describing our primary key as a string. We'll also need to remove the `field :name, :string` line because this is our new primary key. If this seems unusual, recall that the schema doesn't list the `id` field in models where `id` is the primary key.
+Let's move on to `web/models/player.ex` next. We'll need to add a module attribute `@primary_key {:name, :string, []}` describing our primary key as a string. And we need to tell Phoenix how to convert our data structure to an ID that is used in the routes: `@derive {Phoenix.Param, key: :name}` We'll also need to remove the `field :name, :string` line because this is our new primary key. If this seems unusual, recall that the schema doesn't list the `id` field in models where `id` is the primary key.
```
lancehalvorsen: What would you think about changing "And" to "Next" in the newly added section and then add a period at the end? That would look like this:

Then we need to tell . . . routes: `@derive {Phoenix.Param, key: :name}`.

lancehalvorsen:
Hi @iguberman, thanks for submitting this correction! I just left one teeny comment. When we get that one settled and the commits squashed, we'll be all set.

Thanks again!

iguberman:
Hi Lance,

Please feel free to make whatever changes you feel would make this document
better.  Or I'll address your suggestions later tonight -- whatever is
easier for you.  :)

Thank you very much!
Irina.

lancehalvorsen:
Hi @iguberman,

Thanks! I'll merge this and take care of the changes. Thanks so much once again!

iguberman:
No -- thank YOU! :)

PR #378, opened 2015-09-21
Minor fixes in the channels guide

jeffkreeftmeijer:
I set out to update the channels guide, but here are some minor fixes to the current one first.

### DIFF
--- a/G_channels.md
+++ b/G_channels.md
@@ -1,4 +1,4 @@
-Channels are a really exciting and powerful part of Phoenix, one that allows us to easily add soft-realtime features to our applications. Channels are based on a simple idea - sending and receiving messages. Senders broadcast messages about topics. Receivers subscribe to topics so that they can get those messages. Senders and receivers can switch roles on the same topic at any time.
+Channels are a really exciting and powerful part of Phoenix that allow us to easily add soft-realtime features to our applications. Channels are based on a simple idea - sending and receiving messages. Senders broadcast messages about topics. Receivers subscribe to topics so that they can get those messages. Senders and receivers can switch roles on the same topic at any time.
 
 Since Elixir is based on message passing, you may wonder why we need this extra mechanism to send and receive messages. With Channels, neither senders nor receivers have to be Elixir processes. They can be anything that we can teach to communicate over a Channel - a JavaScript client, an iOS app, another Phoenix application, our watch. Also, messages broadcast over a Channel may have many receivers. Elixir processes communicate one to one.
 
@@ -8,11 +8,11 @@ The word "Channel" is really shorthand for a layered system with a number of com
 
 - Socket Handlers
 
-Phoenix holds a single connection to the server and multiplexes your channel sockets over that one connection. Socket handlers, such as your `web/channels/user_socket.ex`, are modules that authenticate and identify a socket connection and allow you to set default socket assigns for use in all channels.
+Phoenix holds a single connection to the server and multiplexes your channel sockets over that one connection. Socket handlers, such as `web/channels/user_socket.ex`, are modules that authenticate and identify a socket connection and allow you to set default socket assigns for use in all channels.
 
 - Channel Routes
 
-These are defined in Socket handlers, such as your `web/channels/user_socket.ex` file, which makes them distinct from other routes. They match on the topic string and dispatch matching requests to the given Channel module. The star character `*` acts as a wildcard matcher, so in the following example route, requests for `sample_topic:pizza` and `sample_topic:oranges` would both be dispatched to the `SampleTopicChannel`.
+These are defined in Socket handlers, such as `web/channels/user_socket.ex`, which makes them distinct from other routes. They match on the topic string and dispatch matching requests to the given Channel module. The star character `*` acts as a wildcard matcher, so in the following example route, requests for `sample_topic:pizza` and `sample_topic:oranges` would both be dispatched to the `SampleTopicChannel`.
 
 ```elixir
 channel "sample_topic:*", HelloPhoenix.SampleTopicChannel
@@ -35,7 +35,7 @@ It is worth noting that these modules are intended for Phoenix's internal use. C
 - Messages
 
 The `Phoenix.Socket.Message` module defines a struct with the following keys which denotes a valid message. From the [Phoenix.Socket.Message docs](http://hexdocs.pm/phoenix/Phoenix.Socket.Message.html).
-  - `topic` - The String topic or topic:subtopic pair namespace, ie “messages”, “messages:123”
+  - `topic` - The String topic or topic:subtopic pair namespace, i.e. “messages”, “messages:123”
   - `event` - The String event name, ie “join”
   - `payload` - The String JSON message payload
   - `ref` - The unique string used for replying to incoming events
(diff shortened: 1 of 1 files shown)


G_channels.md
```diff
@@ -50,7 +50,7 @@ The transport layer is where the rubber meets the road. The `Phoenix.Channel.Tra
 
 - Transport Adapters
 
-The default transport mechanism is via WebSockets which will fall back to LongPolling if WebSockets are not available/working. Other transport adapters are possible, and indeed, we can write our own if we follow the adapter contract. Please see `Phoenix.Transports.WebSocket` for an example.
+The default transport mechanism is via WebSockets which will fall back to LongPolling if WebSockets are not available or not working. Other transport adapters are possible, and we can write our own if we follow the adapter contract. Please see `Phoenix.Transports.WebSocket` for an example.
```
chrismccord: It only falls back if `window.WebSocket` is not defined, or if you explicitly tell it to use LongPoll, so the existing sentence is correct

G_channels.md
```diff
@@ -96,17 +95,16 @@ defmodule HelloPhoenix.RoomChannel do
   def join("rooms:lobby", auth_msg, socket) do
     {:ok, socket}
   end
+
```
lancehalvorsen: It's common to not have extra blank lines between different clauses of the same function, so I think the original is fine here. (Though I would love to know if I'm wrong about this.)

G_channels.md
```diff
@@ -1,4 +1,4 @@
-Channels are a really exciting and powerful part of Phoenix, one that allows us to easily add soft-realtime features to our applications. Channels are based on a simple idea - sending and receiving messages. Senders broadcast messages about topics. Receivers subscribe to topics so that they can get those messages. Senders and receivers can switch roles on the same topic at any time.
+Channels are a really exciting and powerful part of Phoenix that allows us to easily add soft-realtime features to our applications. Channels are based on a simple idea - sending and receiving messages. Senders broadcast messages about topics. Receivers subscribe to topics so that they can get those messages. Senders and receivers can switch roles on the same topic at any time.
```
lancehalvorsen: I think that if we're going to eliminate "one", we'll need to change "allows" to "allow" so the subject and verb will agree.

G_channels.md
```diff
@@ -8,7 +8,7 @@ The word "Channel" is really shorthand for a layered system with a number of com
 
 - Socket Handlers
 
-Phoenix holds a single connection to the server and multiplexes your channel sockets over that one connection. Socket handlers, such as your `web/channels/user_socket.ex`, are modules that authenticate and identify a socket connection and allow you to set default socket assigns for use in all channels.
+Phoenix holds a single connection to the server and multiplexes your channel sockets over that one connection. Socket handlers, such as your `web/channels/user_socket.ex` file, are modules that authenticate and identify a socket connection and allow you to set default socket assigns for use in all channels.
```
lancehalvorsen: What would you think about removing the "your" here as well? That would be:

Socket handlers, such as `web/channels/user_socket.ex`, are . . .

lancehalvorsen:
Hi @jeffkreeftmeijer, thanks for this! There are a few small comments here, so when we get those straightened out and squash the commits, we'll be ready to go.

Thanks again!

G_channels.md
jeffkreeftmeijer: I think that'd be better. I initially wanted to do that, until I noticed that "your `path/to/file`" was used a couple of times throughout the guide.

G_channels.md
jeffkreeftmeijer: I could be wrong, but I think it's correct like this, as "part" is the subject here, not "channels". 

"(Channels are) a powerful part of Phoenix that allows us to ..."`

G_channels.md
lancehalvorsen: Actually, if you would be interested :), either removing the instances of "your" or converting them to a more neutral "the" where applicable would be a nice public service.

G_channels.md
lancehalvorsen: Interesting. I read the sentence as, "Channels allow us . . . " with "Channels" as the subject and "allow" as the verb. I also read the "are a powerful part of Phoenix that" as an adjectival clause describing "Channels".

Does that make sense, or should I write an angry letter to my 8th grade English teacher? ;^) ;^) ;^)

G_channels.md
jeffkreeftmeijer: Sure, I found three more in the rest of the guides; https://github.com/phoenixframework/phoenix_guides/pull/379

G_channels.md
jeffkreeftmeijer: Ah, you got me. I'm off writing an angry letter to mine. ;)

lancehalvorsen:
:tada: Thanks @jeffkreeftmeijer! :tada:

chowza:
In the channels guide in the `app.html` section I needed to include `<script>require("web/static/js/app")</script>` before the closing body tag to make this section work. 

You might want to double check that is the case on your end as well and that isn't just my problem. @lancehalvorsen @jeffkreeftmeijer

jeffkreeftmeijer:
@chowza do you mean in `web/templates/layout/app.html.eex`? I have `<script src="<%= static_path(@conn, "/js/app.js") %>"></script>` already there after generating a new Phoenix project. Could you push your app to Github and open a new issue so I can take a look?

chowza:
Yea `web/templates/layout/app.html.eex` is what I meant. I also have the static_path there, and when I go to my `localhost:4000` I see a script tag with `src='app.js?vsn=6782'` (vsn number changes with live reloading). In any case without having `<script>require("web/static/js/app")</script>` it doesn't seem like anything in app.js actually executes. A simple test would be to create a new Phoenix app. Go to app.js and type `console.log("this log is my app js")`. Mix phoenix.server and go to localhost:4000. There will be no console logs. Now add the `<script>require("web/static/js/app")</script>`. You will see the console log now. I can push something to git later tonight when I get home from work, but I don't think it's necessary

jeffkreeftmeijer:
I've created a test app to make sure we're talking about the same thing; https://github.com/jeffkreeftmeijer/phoenix_guides-issue-378

It's a Phoenix 1.0.2 app with an `alert()` in `web/static/js/app.js`, which fires when I visit http://localhost:4000. One change from your explanation is that my script tag's source seems to be different from yours;

    <script src="/js/app.js?vsn=76C0D56"></script>

Could you try if my example app works for you and open a new issue https://github.com/phoenixframework/phoenix/issues if it doesn't? If that's the case, this sounds more like a framework bug than an error in the documentation to me.

chowza:
@jeffkreeftmeijer I cloned your app and ran the server `mix phoenix.server` and when I went to `localhost:4000` I didn't get any alerts. Putting `<script>require("web/static/js/app")</script>` above the closing body tag made me get the alert. My phoenix is 1.02 as well, running on a windows 8.1. If you are saying it works on yours then this isn't a guide issue and must be some other issue. 


PS `app.js?vsn=12345` will differ from other ppls during development because that is used for updating the app.js immediately. If you for example update app.js while your server is running you should see the numbers change. In production that script will become simply `app.js`

jeffkreeftmeijer:
Yes, it's working for me, so this sounds like a bug in Phoenix. Please open a new issue in https://github.com/phoenixframework/phoenix/issues and link to this issue and the example app from there, so we can figure this out.

Also, I know that the `vsn` parameter changes for live reloading, but I was talking about the path to `app.js`. You told me your script tag looks like `<script src="app.js"></script>`, while mine is `<script src="/js/app.js"></script>` (I removed the `vsn` attributes for clarity). That could be the problem.

chrismccord:
Brunch has an `autoRequire` feature that we started using several versions back. This removed the need for `<script>require("web/static/js/app")</script>`, but you need to enable it in your brunch-config and make sure you are running the proper version of brunch. See here for an example:

https://github.com/phoenixframework/phoenix/blob/master/installer/templates/static/brunch/brunch-config.js#L60-L64

chowza:
@chrismccord that makes sense. I also had <a href="http://stackoverflow.com/questions/30147801/phoenix-framework-invalid-argument-at-new-socket-windows">this problem </a> with brunch, and obviously if brunch was not working the autoRequire was also not working. Thanks!

PR #380, opened 2015-09-22
Add intercept in RoomChannel example

ChristopheBelpaire:
```handle_out``` is never called in the example without ```intercept ["new_msg"]```

### DIFF
--- a/G_channels.md
+++ b/G_channels.md
@@ -194,7 +194,8 @@ We handle incoming events with `handle_in/3`. We can pattern match on the event 
 ```elixir
 defmodule HelloPhoenix.RoomChannel do
   use Phoenix.Channel
-
+  intercept ["new_msg"]
+  
   def join("rooms:lobby", auth_msg, socket) do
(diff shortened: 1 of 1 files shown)


lancehalvorsen:
Hi @ChristopheBelpaire, thanks for contributing this! Sorry I was delayed getting to it.

In this case, I don't believe we need to intercept the `new_msg` event. We only want to do so if we need to filter the output we broadcast for that event in some way. For `new_msg`, we simply broadcast to everyone without any filter.

Please see the next section in the Channels Guide, "Intercepting Outgoing Events" for more info. Please also see the Phoenix docs. http://hexdocs.pm/phoenix/Phoenix.Channel.html#intercept/1 

Thanks again!

ChristopheBelpaire:
Hello @lancehalvorsen,
It's just that I copied/pasted the code, and I didn't understood why the handle_out function wasn't called until I read the next paragraph about Intercepting Outgoing Events :)

PR #384, opened 2015-10-01
Clarifications to the Heroku guide

jeffkreeftmeijer:
I tried to clarify the Heroku guide a bit.

- Don't start by copying the contents of `config/prod.secrets.exs` to `config/prod.exc` and then completely rewrite it, but simply add the required lines in one go.
- Put the `secret_key_base`-configuration in the existing config block
- Don't use "A_LONG_STRING_WILL_BE_PRINTED"/"my_secret_key_base", but use an example key
- Write a nicer commit message

Here's a demo app of the guide's exact end result (to update the guide with later): https://github.com/jeffkreeftmeijer/phoenix_heroku_example


lancehalvorsen:
Thanks so much for this @jeffkreeftmeijer! I really appreciate your contribution. <3

Sorry it took me a while to get to it.

PR #385, opened 2015-10-02
Fixed a few typos and updated views guide

wiserfirst:
1 Fixed a few typos: were -> was
2 Updated the html in the Views guide to match the latest Phoenix version (1.0.3)
3 Updated a few words in the Views guide

### DIFF
--- a/D_controllers.md
+++ b/D_controllers.md
@@ -420,7 +420,7 @@ For a list of valid content mime-types, please see the [mime.types](https://gith
 
 We can also set the HTTP status code of a response similarly to the way we set the content type. The `Plug.Conn` module, imported into all controllers, has a `put_status/2` function to do this.
 
-`put_status/2` takes `conn` as the first parameter and as the second parameter either an integer or a "friendly name" used as an atom for the status code we want to set. Here is the list of supported [friendly names](https://github.com/elixir-lang/plug/blob/master/lib/plug/conn/status.ex#L7-L63).
+`put_status/2` takes `conn` as the first parameter and as the second parameter either an integer or a "friendly name" used as an atom for the status code we want to set. Here is the list of supported [friendly names](https://github.com/elixir-lang/plug/blob/master/lib/plug/conn/status.ex#L7-L66).
 
 Let's change the status in our `PageController` `index` action.
 
--- a/E_views.md
+++ b/E_views.md
@@ -16,7 +16,7 @@ end
 
 That's simple enough. There's only one line, `use HelloPhoenix.Web, :view`. This line calls the `view/0` function we just saw above. Besides allowing us to change our template root, `view/0` exercises the `__using__` macro in the `Phoenix.View` module. It also handles any module imports or aliases our application's view modules might need.
 
-At the top of this file, we mentioned that views are a place to put functions for use in our templates. Let's experiment with that a little bit.
+At the top of this guide, we mentioned that views are a place to put functions for use in our templates. Let's experiment with that a little bit.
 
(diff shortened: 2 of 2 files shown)


B_adding_pages.md
```diff
@@ -237,7 +237,7 @@ It's good to remember that the keys to the params [Dict](http://elixir-lang.org/
 
 For the last piece of this puzzle, we'll need a new template. Since it is for the `show` action of the `HelloController`, it will go into the `web/templates/hello` directory and be called `show.html.eex`. It will look surprisingly like our `index.html.eex` template, except that we will need to display the name of our messenger.
 
-To do that, we'll use the special eex tags for executing Elixir expressions - `<%=  %>`. Notice that the initial tag has an equals sign like this: `<%=` . That means that any Elixir code that goes between those tags will be executed, and the resulting value will replace the tag. If the equals sign were missing, the code would still be executed, but the value would not appear on the page.
+To do that, we'll use the special eex tags for executing Elixir expressions - `<%=  %>`. Notice that the initial tag has an equals sign like this: `<%=` . That means that any Elixir code that goes between those tags will be executed, and the resulting value will replace the tag. If the equals sign was missing, the code would still be executed, but the value would not appear on the page.
```
lancehalvorsen: I believe that "were" here is correct. It's the subjunctive case.

C_routing.md
```diff
@@ -46,7 +46,7 @@ Inside the scope block, however, we have our first actual route.
 
 The first argument to these macros is the path. Here, it is the root of the application, `/`. The next two arguments are the controller and action we want to have handle this request. These macros may also take other options, which we will see throughout the rest of this guide.
 
-If this were the only route in our router module, the clause of the `match/3` function would look like this after the macro is expanded.
+If this was the only route in our router module, the clause of the `match/3` function would look like this after the macro is expanded.
```
lancehalvorsen: I believe that "were" is correct here as well - also the subjunctive case.

deployment/I_exrm_releases.md
```diff
@@ -312,7 +312,7 @@ $ rel/hello_phoenix/bin/hello_phoenix ping
 pong
 ```
 
-It's still alive and responding. This is a feature! If the server were to go down every time we stopped a remote console, that would be a problem. Note that this situation is different from when we ran the `console` command above. The `remote_console` command is not intended to start a release locally, but rather to connect to one which has already started.
+It's still alive and responding. This is a feature! If the server was to go down every time we stopped a remote console, that would be a problem. Note that this situation is different from when we ran the `console` command above. The `remote_console` command is not intended to start a release locally, but rather to connect to one which has already started.
```
lancehalvorsen: This is another example of the subjunctive case.

lancehalvorsen:
Hi @wiserfirst. Thanks so much for this contribution. Sorry I was so late getting to it. ElixirConf and a killer cold have held me up for a few days.

I've left a few very small comments. If we can address those and squash the commits, we'll be ready to go.

Thanks again!

maxbeizer:
+1 to the [subjunctive case](https://en.wikipedia.org/wiki/English_subjunctive) comments.

wiserfirst:
@lancehalvorsen Thank you very much for you comments. Sorry for my bad grammar.

I've reverted all the updates for were to was and squashed the commits. Please review.

Thanks again!

lancehalvorsen:
Hi @wiserfirst, this is great! There's no need to apologize at all. The subjunctive case is a little unusual.

Thanks so much for your contribution. <3 <3 <3

wiserfirst:
@lancehalvorsen Really glad I can contribute :satisfied:

PR #387, opened 2015-10-03
Added correct Heroku url

KamilLelonek:
All Heroku applications are hosted under `herokuapp` domain so each of them looks like `app-name.herokuapp.com`. [heroku.com](http://heroku.com/) is only for accessing a dashboard and no user apps are hosted as its subdomain.

[Review on Reviewable](https://reviewable.io/reviews/phoenixframework/phoenix_guides/387)


lancehalvorsen:
Good catch @KamilLelonek, thank you! <3

PR #388, opened 2015-10-04
Fixes to the testing introduction guide

jeffkreeftmeijer:
* Remove the "==> ranch (compile)" and  "Generated hello_phoenix app"
lines and the three dots from the first test run output to avoid
confusion with the four dots for the running tests. Now the output
matches the output you'd get after having compiled before, like the rest
of the test output in the guide.
* Add `test/views/layout_view_test.exs` to the `tree test` output
* Remove the "Create the database..."-comment in test/test_helper.exs,
as it's been removed in
https://github.com/phoenixframework/phoenix/commit/fcdc47a5ae19c00fe00ccb83c717c3c95af8685e.
* Add `end` tags to the `test/views/error_view_test.exs` code blocks
* Include the full output for `$ mix test
test/views/error_view_test.exs:12`
* Add a `$` before `mix test --exclude error_view_case`
* Include the `import Phoenix.View`-line and match indentation to the
generated file in `test/views/error_view_test.exs`
* Remove double "value" in "We can also specify a value and run only
tests with that (value)."

Generated using https://github.com/jeffkreeftmeijer/phoenix_testing_introduction_example/blob/master/README.md.erb to make sure the file contents and console output is correct. 

I've gotten to the "Generating More Files" heading and plan on finishing the whole guide, but I'm sending this PR now because I don't know if I'll be able to get to it soon.

### DIFF
--- a/testing/A_introduction.md
+++ b/testing/A_introduction.md
@@ -12,11 +12,7 @@ In a freshly generated application, let's run `mix test` at the root of the proj
 
 ```console
 $ mix test
-==> ranch (compile)
-
-. . .
-
-Generated hello_phoenix app
+....
 
 Finished in 0.2 seconds (0.2s on load, 0.00s on tests)
 4 tests, 0 failures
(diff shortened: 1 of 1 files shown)


testing/A_introduction.md
```diff
@@ -41,10 +37,11 @@ test
 ├── test_helper.exs
 └── views
     ├── error_view_test.exs
+    ├── layout_view_test.exs
```
lancehalvorsen: This is funny. I actually wrote the PR in Phoenix to put this file in the generated application, but I didn't fix the guide. Derp. ;^)

Thanks for catching my mistake!

lancehalvorsen:
Thanks for this contribution @jeffkreeftmeijer! <3 <3 <3

PR #389, opened 2015-10-06
Add acute accent to Fred Hébert's surname

maxbeizer:
"Hebert" should be "Hébert" according to the [cover of Learn You Some Erlang](http://www.amazon.com/Learn-Some-Erlang-Great-Good/dp/1593274351).

:heart: fellow Herokai.


lancehalvorsen:
This is a great catch @maxbeizer. Thank you so much! <3

PR #390, opened 2015-10-06
Update link to Heroku sign up

maxbeizer:
Though `/www-header` works, it's actually an artifact of legacy Heroku code. This update makes the link more clear by removing the path.

source: I work on the Heroku www & signup apps.


lancehalvorsen:
Thanks again @maxbeizer! I certainly trust your Heroku cred on this one! :^) <3 <3 <3

PR #393, opened 2015-10-08
Fixed typo in Testing Models guide

wiserfirst:
1. In Testing Models guide, the string length returned by long_string/1 function is `length+1`, which is probably not intended.
2. Updated some wording in Static Assets Guide.

And also the last section of Testing Models Guide, `Generating an HTML Resource` has already been covered in Ecto Models Guide. If the repetition is unnecessary, I can remove this section as well. Thanks!


lancehalvorsen:
Thanks @wiserfirst! <3 <3 <3

PR #398, opened 2015-10-11
Fix the code block got included in quote

tianyi33:
Very small issue, , the code block got included in the `<` quote,  which also affects normal text rendering at the next line. 

before:
before

after:
after


lancehalvorsen:
Thanks so much for this contribution @tianyi33! <3

PR #400, opened 2015-10-12
Fix default value syntax

hubertlepicki:


lancehalvorsen:
Nice catch @hubertlepicki! Thanks so much for this. <3 <3 <3

PR #403, opened 2015-10-13
Update example tree of default app.

JeroenKnoops:
A new phoenix project does not use scss, but uses css.

`app.scss` --> `app.css`

### DIFF
--- a/B_adding_pages.md
+++ b/B_adding_pages.md
@@ -48,7 +48,7 @@ priv
 web
 └── static
     ├── css
-    |   └── app.scss
+    |   └── app.css
     ├── js
     │   └── app.js
     └── vendor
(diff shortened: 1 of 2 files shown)


jeffkreeftmeijer:
Hey @JeroenKnoops, nice find! There's another one in https://github.com/phoenixframework/phoenix_guides/blob/cd81b655f85e6d6b954e0aae4a9ce42f421f68b9/bonus_guides/D_mix_tasks.md#L444, could you update that too and squash? Just so we have fixed this once and for all. :)

lancehalvorsen:
@JeroenKnoops, thanks for submitting this! I agree with @jeffkreeftmeijer, if you could change the other instance of this error in this PR, that would be really nice.

Thanks!

JeroenKnoops:
Updated the update! :)

lancehalvorsen:
Awesome! Thanks @JeroenKnoops! <3 <3 <3

PR #405, opened 2015-10-15
change 'use' line of session example

y-yagi:
It was corrected so as to be the same as the file actually is generated.


lancehalvorsen:
Thanks so much for catching this @y-yagi! That's a really old one. :)  <3 <3 <3

y-yagi:
Thanks!

PR #406, opened 2015-10-16
Update the model testing guide

jeffkreeftmeijer:
I've updated the models testing guide to make sure file contents and console output is accurate. I haven't checked the "Generating an HTML Resource"-section yet (I need to build another example app for that), but I hope to get to that soon.

* Remove compile and generation lines. Now the output matches the output
  you'd get after having compiled before, like the test output in the
  testing introduction guide.
* Update the test counts. The testing introduction guide leaves the user
  with an app that has sixteen tests now, so this guide is working from
  there.
* Add module lines around code blocks (like `defmodule HelloPhoenix.UserTest do`
  and `end`) to add context and indent accordingly.
* Add the changes in UserController as an explicit step instead of a
  note to prevent test failures later in the guide.
* Display the schema information in the `User` model code block to add
  context and not have to break the file up twice.
* Update the test output to ExUnit 1.1.1. Uses "1 failure" instead of "1
  failures", displays the amount of skipped tests in the output, and has
  nicer indentation in the test results.

Here's the example app I used for both of the testing guides; https://github.com/jeffkreeftmeijer/phoenix_testing_example

### DIFF
--- a/testing/E_models.md
+++ b/testing/E_models.md
@@ -6,18 +6,15 @@ Before we do anything else, let's run `mix test` to make sure our test suite run
 
 ```console
 $ mix test
-Compiled lib/hello_phoenix.ex
-. . .
-Generated hello_phoenix app
-...............
+................
 
 Finished in 0.6 seconds (0.5s on load, 0.1s on tests)
-15 tests, 0 failures
+16 tests, 0 failures
 
 Randomized with seed 638414
 ```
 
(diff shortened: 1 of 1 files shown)


lancehalvorsen:
This is fabulous @jeffkreeftmeijer! Thanks once again for your contribution! <3 <3 <3

PR #407, opened 2015-10-17
fix missing comma in email config

datachomp:
The missing comma creates the following error:
```
** (Mix.Config.LoadError) could not load config config/config.exs
    ** (SyntaxError) config/config.exs:29: syntax error before: mailgun_key
```
This pull request addresses a minor problem in the "sending email" documentation that can be a hurdle to newer people to phoenix as it is subtle and completely breaks a persons phoenix app.


lancehalvorsen:
Good catch @datachomp! Thanks very much. <3

PR #408, opened 2015-10-17
fix URL of elixir guides

y-yagi:
I think now that it http://elixir-lang.org/getting-started/introduction.html  is correct URL


lancehalvorsen:
Nice! Thanks for catching this, @y-yagi. <3

y-yagi:
Thanks!

PR #409, opened 2015-10-17
Update broken and 301'-ing links

jeffkreeftmeijer:
After seeing #408, I used [this little script](https://gist.github.com/jeffkreeftmeijer/5e36a0cfa7e2c735386e) to check the rest of the urls in the guides. 

* ~~Link to the sign up page from the email guide, and replace occurences
  of "mailgun.com" with "Mailgun".~~ moved to https://github.com/phoenixframework/phoenix_guides/pull/411
* Use https links to pragprog.om, manning.com and nostarch.com
* Update links to https://www.nginx.com
* Use confreaks.tv instead of .com
* Use elements.heroku.com/addons instead of addons.heroku.com
* Switch to hexdocs.pm for exrm's documentation

PR #411, opened 2015-10-18
Differentiate between "Mailgun" and `mailgun`

jeffkreeftmeijer:
- Link to the sign up page from the email guide
- Use "Mailgun" to refer to the service
- Use `mailgun` to refer to the package
- Use `:mailgun_domain` and `:mailgun_key` to refer to `mailgun`'s configuration options

From @lancehalvorsen in https://github.com/phoenixframework/phoenix_guides/pull/409#issuecomment-148973105:
> Originally, I had used "mailgun.com" to make it painfully obvious when we were talking about the service rather than "mailgun" the package. Do you (or anybody else who might be watching) feel that using "Mailgun" for the service and "mailgun" for the package is clear enough?

I think using "Mailgun" to refer to the service and `maligun` to refer to the package is clear enough, especially if we wrap `maligun` in `<code>` tags.


lancehalvorsen:
Very nice. Thank you! <3 <3 <3

PR #412, opened 2015-10-18
Update the "Generating .." section in the models testing guide

jeffkreeftmeijer:
As promised in #406;

- Update output for `mix.gen` to Ecto 1.0.6.
- Display the comment at the end of `web/router.ex` to add context and
  not have to break the file up twice.
- Remove a double space.


lancehalvorsen:
I'm running out of nice ways to that Thank You @jeffkreeftmeijer, but that's a great problem to have. :)  I really appreciate all the work you've put in on the guides. <3 <3 <3

PR #413, opened 2015-10-18
remove dashes and update phoenix version to 1.0.3

Kosmas:


lancehalvorsen:
Very nice @Kosmas. Thanks very much! <3

PR #419, opened 2015-10-28
Describe how to install & import JS libs with NPM

eikes:


bonus_guides/G_static_assets.md
```diff
@@ -148,6 +148,10 @@ Reload the page. Open the JS console and type `meaning_of_life`. This will retur
 
 Important detail: according to the default configuration there is no ES6 support for files in `web/static/vendor`. Should we need to enable it, look for `plugins: { babel: { ignore:` in `brunch-config.js`.
 
+#### JavaScript Libraries
+
+If you want to use a JS library like jQuery, underscore or whatever you need, you could copy it to the `web/static/vendor` folder, as mentioned above. It may be a little bit easier to use `npm` to install it. Simply add `"jquery": ">= 2.1"` to the dependencies in the `package.json` file in your projects root and run `npm install --save`. Now you can `import $ from "jquery"` in your `app.js`.
```
lancehalvorsen: One thing we try to do in the guides is minimize the use of "you" and "your". We tend to substitute "we" and "our" or remove the possessive pronoun altogether. The idea is to foster a subtle sense of inclusion with the reader, as if to say  that we're all in this programming thing together. :)

There are a number of instances of "you" and "your" in this paragraph. Let me take the first sentence or two as an example of how to turn this around:

We may need to use a JavaScript library like jQuery or underscore in our application. As we mentioned above, we could copy the libraries into `web/static/vendor`.

lancehalvorsen:
Hi @eikes! Thanks very much for taking the time to write up this PR. <3

I left a note about usage. It would be great if you could apply that idea to the rest of the paragraph.

Thanks again.

eikes:
Hi @lancehalvorsen!

I did change the words, to match the guide's style. On a more technical note: I noticed that when you install jQuery like this, it becomes modularized, ie. not in the global scope, so it's only available after explicityl importing it in a JavaScript file. That may not be what we want. Is there a way to install the NPM installed version into the global scope?

lancehalvorsen:
Hi @eikes, so sorry for the delay. I want to check in with Chris about this one, and we haven't had a chance to connect yet. We appreciate your patience on this!

chrismccord:
:shipit:

PR #420, opened 2015-10-29
Use GenServer vs. Genserver

scrogson:


lancehalvorsen:
Nice catch. Thanks so much @scrogson! <3 <3 <3

PR #422, opened 2015-11-01
Update the templates guide

jeffkreeftmeijer:
I've updated the templates testing guide to make sure the code blocks are clear and accurate, fix some typos and do some styling fixes and in line with the rest of the guides:

* Change "…what they […] should be - files…" to "…what they […] should
be: files…"
* "…generates `web/web.ex` file…" => "…generates a `web/web.ex file…"
* Explicitly state that "the main view" is in `web/web.ex`
* Add module lines around code blocks (like `defmodule HelloPhoenix.PageController do`
  and `end`) to add context and indent accordingly.
* Display the comment at the end of web/router.ex to add context and not have to break the file up twice.
* Use `html` instead of `elixir` as the info string for template code blocks.
* Remove trailing newlines in code blocks
* Add missing newlines after code blocks
* Remove a double newline before the “Displaying Lists” heading

Here's the example app I used: https://github.com/jeffkreeftmeijer/phoenix_guides-templates-example


lancehalvorsen:
Thanks for cleaning up some crufty old stuff, @jeffkreeftmeijer! I really appreciate your contributions! <3 <3 <3

PR #423, opened 2015-11-01
Don't explain how to change the template root directory

jeffkreeftmeijer:
While interesting to know, explaining how to change the template root directory is out of the scope of the templates guide, as changing the template root is something you should probably only do when you're more familiar with the framework. Also, putting it at the start of the guide like this is confusing, as it's not completely clear if the reader should do anything with the template root directory or not. It might be
a good idea to move this elsewhere.

I'm opening a separate pull request for this instead of throwing it in with #422 to have some room for discussion wether this should be removed or not.


lancehalvorsen:
Hi @jeffkreeftmeijer thanks for this, and sorry for the delay. I wanted to mull this over a little. I think you're right. Since the View Guide covers changing the template root directory pretty thoroughly, and since we link to the View Guide, it's not imperative to cover it here.

Thanks again!

PR #424, opened 2015-11-03
Minor fixes to the templates guide

jeffkreeftmeijer:
This seems to have slipped through when working on #422.

* Mention the filename (`web/templates/page/test.html.eex`) in the
  paragraph above the last code block, as moving one file and editing
  another might be confusing
* Use `html` instead of `elixir` as the info string for the last code
  block

### DIFF
--- a/F_templates.md
+++ b/F_templates.md
@@ -183,9 +183,9 @@ defmodule HelloPhoenix.SharedView do
 end
 ```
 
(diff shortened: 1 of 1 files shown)


lancehalvorsen:
Nice catches once again, @jeffkreeftmeijer. Thanks! <3 <3 <3

PR #425, opened 2015-11-07
Clarifying a hiccup in the guide when using --no-brunch

landongn:
There's an implied notion in the channel guide that you've created a test project with brunch.  When I wasn't able to find the related files the guide spoke of, I didn't have the context that they only came along with the brunch-enabled generator.

This adds a straightforward sentence to the guide that explains that these files are present only in the 'standard' generator, so users have an inkling on why they might not have the same files.  True story. 

Also, bumped jQuery to use the latest  < 1.x version (as of now, 1.11.9)


jeregrine:
Thanks!

PR #426, opened 2015-11-07
Added NGINX config for reverse proxy sockets

landongn:
Added some information to the "Phoenix Behind A Proxy" guide, which shows a standard reverse proxy setup ensuring forwarded headers are present.


jeregrine:
@landongn this is great! but its hard for us to verify and test this :( which explains the long delay.

PR #428, opened 2015-11-11
Fix case to match project docs: Eex -> EEx

cmoel:
I noticed http://elixir-lang.org/docs/stable/eex/EEx.html uses the EEx case while the guides had it as Eex


lancehalvorsen:
Thanks for taking the time to create a PR @cmoel! Sorry for the delay in getting to this. 

Much appreciated. <3

PR #429, opened 2015-11-16
added missing static dir to web directory structure example

fblecha:


lancehalvorsen:
Hi @fblecha, thanks for your patience and you contribution! <3

PR #431, opened 2015-11-23
Update the "Up and running" guide

jeffkreeftmeijer:
* Use "let's run `mix phoenix.new`" instead of "let's call
  `phoenix.new`"
* Show more output from the `mix phoenix.new` task and include the
  `Fetch and install dependencies? [Yn]`-line
* Split the output from `mix phoenix.new` only once, at the "Fetch and
  install dependencies"-prompt
* Split up `cd`-ing into the project directory, creating the database
  and starting the Phoenix server and include output
* Update output from `mix phoenix.server`


lancehalvorsen:
Thanks so much @jeffkreeftmeijer and sorry for the delay in merging this!

<3 <3 <3

PR #432, opened 2015-11-28
Added instructions for non default postgres passwords

efexen:
Hey,

I was going through the Up and Running guide and ran into a problem with my non default postgres role password. Having a look in the ecto.create specific part of the documentation I found a whole host of error messages and their resolutions so I thought this might be a good addition?

I've attempted to keep to language similar to that used on description of the other issues.

Thanks,
.FxN


lancehalvorsen:
Thanks @efexen! <3 <3 <3

PR #433, opened 2015-11-29
custom_errors: clarify example

henrik:
Confused by the fact that you don't see the error page in dev (I think I'll make a separate PR about that), I added this handling in myself and got an "redefining" warning.


lancehalvorsen:
Thanks @henrik! <3 <3 <3

PR #437, opened 2015-12-08
Fix indetion in `handler_info` code

antage:


lancehalvorsen:
Thanks for this @antage! <3 <3 <3

PR #441, opened 2015-12-15
Fix incorrect/missing code in find_message plug

dmarkow:
I was using the Plug guide as a reference and it appeared that `params` would be available in my plug based on this guide. I then got a `function params/0 undefined` error. Changing my plug to reference `conn.params` instead gave me what I was looking for.

There was also a missing `do` at the end of one of the `case` statements.


lancehalvorsen:
Thanks @dmarkow! <3 <3 <3

PR #444, opened 2015-12-18
tip to prevent issues on debian based distros

erikdsi:
little tip to prevent https://github.com/phoenixframework/phoenix/issues/1411


lancehalvorsen:
Thanks very much @erikdsi! <3 <3 <3

PR #448, opened 2015-12-26
Upgrade section of guides

radar:
Rather than having the (rather excellent) upgrade guides hidden away in Gists by @chrismccord, I think they should be made a prominent feature of the guides (similar to how Rails has an [upgrading guide](http://guides.rubyonrails.org/upgrading_ruby_on_rails.html)).

I've added an example of one of those guides here and want to open this PR and discussion before I go doing any more work on it.

### DIFF
--- a/upgrading/A_0-16-to-0-17.md
+++ b/upgrading/A_0-16-to-0-17.md
@@ -1,0 +1,67 @@
+This guide will walk you through the steps to upgrade your Phoenix application from 0.16.x to 0.17.x.
+
+## Deps
+
+Bump your `:phoenix` and `:phoenix_live_reload` deps in `mix.exs`:
+
+```elixir
+def deps do
+  [...
+   {:phoenix, "~> 0.17"},
+   {:phoenix_live_reload, "~> 1.0"},
+  ...]
+end
+```
+
+## Endpoint
+
+The `:render_errors: [default_format: "html"]` is deprecated in favor of `render_errors: [accepts: ["html"]]`. Update your deprecated references in `config/*.exs`
+
+`plug :router` is no longer needed and has been removed. Update your `plug :router, Myapp.Router` to `plug Myapp.Router`.
+
+## Controller
+
+The "format" param for overriding the accept header has been renamed to "\_format" and is no longer injected into the params when parsing the Accept headers. Use `get_format/1` to access the negotiated format. If you were relying on the format param being present to match in the controller, such as:
+
+```elixir
+def show(conn, %{"id" => id, "format" => "html"}) do ...
+```
+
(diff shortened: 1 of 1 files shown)


lancehalvorsen:
Thanks @radar for taking the time to create this PR! Here's the history on the Upgrade Guides.

There used to be an individual guide for each upgrade, eg. 0.11.0 to 0.12.0. They were all located in the Bonus Guides section. As you can imagine, they began to pile up and obscure the other guides. What we did to fix this is to create a single Upgrade Guide in the Bonus Guides section that points to the new home of all the Upgrade Guides, the News section.

Here's the Upgrade Guide: http://www.phoenixframework.org/docs/upgrading-phoenix
Here's the News section that it points to: http://www.phoenixframework.org/blog

Since the Phoenix changelog is the source of truth, we didn't duplicate them here in the repo.

I will have time this evening and tomorrow to work on some long overdue PRs here. I'll take care of creating a new News item for this.

radar:
Hey @lancehalvorsen, no pressure :)

The changelogs are a great list of all the things that changed in the bit in between versions. What I'm asking for is a clear + concise upgrade guide between versions; essentially the same as @chrismccord's Gists now, but more... uh, "official". 

Would this be something that we could accomplish?

lancehalvorsen:
Hi @radar, 
Sorry, I got lost in my sense of guilt for neglecting PRs and didn't communicate clearly. ;)

We already have this on the phoenixframework.org site. Here's the same content as the PR: http://www.phoenixframework.org/blog/upgrading-from-v016-to-v017

The thing is, we don't have files for these in this repo because they would just be a copy of @chrismccord's changelog.

To address the missing upgrade content, I just need to go into the admin section of readme.io, which currently hosts the guides, and create a new blog item using the changelog markdown for the upgrade from v1.0 to v1.1.

Did I do a better job this time? :) Please let me know if I'm missing something. Thanks again!

radar:
Yes, that's good :) 

Having the blog posts for 0.17 -> 1.0, and 1.0 -> 1.1 would be great :0

lancehalvorsen:
Awesome! 

Actually, there were no external facing changes for the 0.17 -> 1.0 upgrade, iirc, so there were no upgrade instructions.

The 1.0 -> 1.1 guide will happen this weekend.

lancehalvorsen:
This is now live on the site. Unfortunately, readme.io no longer seems to let me drag and drop the items to set their order, so this new one obscures The Road to 2 Million Websocket Connections in Phoenix.

radar:
Thanks a lot for doing this @lancehalvorsen! :+1:

lancehalvorsen:
:+1:

PR #449, opened 2015-12-28
Remove unnecessary double quotes

BanzaiMan:
These double quotes should not be in the file content.

### DIFF
--- a/D_controllers.md
+++ b/D_controllers.md
@@ -354,7 +354,7 @@ end
 What it doesn't have is an alternative template for rendering text. Let's add one at `/web/templates/page/index.text.eex`. Here is our example `index.text.eex` template.
 
 ```elixir
-"OMG, this is actually some text."
+OMG, this is actually some text.
 ```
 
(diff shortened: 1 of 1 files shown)


lancehalvorsen:
Hi @BanzaiMan,

Thank you for taking the time to put together this PR!

Since both of these examples are inside of Elixir code blocks, they need to be valid Elixir code. The way to do that is to make them Elixir strings, so the double quotes are necessary.

Thanks again.

BanzaiMan:
If that is the case, these snippets should not be marked as Elixir code segments.

D_controllers.md
```diff
@@ -393,7 +393,7 @@ end
 And let's add a bit to our text template.
 
 ```elixir
-"OMG, this is actually some text." <%= @message %>
+OMG, this is actually some text. <%= @message %>
 ```
 
 Now if we go to `http://localhost:4000/?_format=text&message=CrazyTown`, we will see "OMG, this is actually some text. CrazyTown"
```
BanzaiMan: This output is incorrect. The correct output would be:

```
"OMG, this is actually some text." CrazyTown
```

lancehalvorsen:
Before the year is out, I must say that you are correct! Thank you for persevering, and I very much appreciate this PR. Happy New Year! :tada:

BanzaiMan:
@lancehalvorsen No worries. Thanks for merging! Happy New Year!

PR #450, opened 2015-12-29
Minor typo, should likely read "common imports"

willykaram:
This is a suggested correction to what appears to be a minor typo in the Templates Guide. The current wording is "commons imports", which should likely read "common imports".


lancehalvorsen:
Thank you very much @willykaram! <3

PR #451, opened 2015-12-30
Mention to restart our server after modifying config

willykaram:
Currently the Views Guide doesn't instruct the reader to restart the server after editing the config/dev.exs file. I think this leads to the error in the attached screenshot. Perhaps we should indicate that the server needs to be restarted in order for the config change to take effect or alternatively use this an opportunity to explain this principle/error message
config-error

.

### DIFF
--- a/E_views.md
+++ b/E_views.md
@@ -198,7 +198,7 @@ config :hello_phoenix, HelloPhoenix.Endpoint,
   . . .
 ```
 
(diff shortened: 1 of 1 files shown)


lancehalvorsen:
Thanks very much @willykaram! I think that pointing out the need to restart the server is preferable. The error message should be self-explanatory, I think.

Thanks again!

Issue #439, opened 2015-12-10
Do not use nested scopes in examples


chrismccord:
I push folks away from nesting scopes because it gets confusing and is less clear. We should update the routing code to change this block:

```elixir
  scope "/", HelloPhoenix do
    pipe_through :browser

    resources "posts", PostController

    scope "/reviews" do
      pipe_through :review_checks

      resources "reviews", ReviewController
    end
  end
```

To this:

```elixir
  scope "/", HelloPhoenix do
    pipe_through :browser

    resources "posts", PostController
  end

  scope "/reviews", HelloPhoenix do
    pipe_through [:browser, :review_checks]

    resources "reviews", ReviewController
  end
```

Prose will also need tweaked talking about nested scopes.

willykaram:
@chrismccord @lancehalvorsen I've made a PR to address the above issue. Pls LMK if you have any feedback, and I'll revise accordingly.

willykaram:
As mentioned in the PR, apologies I made a sloppy error while squashing and then another error in my prior branch, and given the time of day it was just easier to create a new PR than to clean-up the error.

lancehalvorsen:
closed by #455

PR #452, opened 2015-12-31
Adjusting nested scopes examples in Routing Guide

willykaram:
Per issue #439, I've removed the nested scope from the example by modifying the code block and related prose mentioned.

I've also adjusted the prose earlier in the Guide that references a nested scope, which isn't as confusing as that mentioned by @chrismccord. That said, we could also potentially eliminate that example too. If so, LMK and I'll make that revision by removing the related prose/code block.

### DIFF
--- a/C_routing.md
+++ b/C_routing.md
@@ -498,7 +498,7 @@ review_path  DELETE  /reviews/:id       HelloPhoenix.ReviewController :delete
   user_path  DELETE  /users/:id         HelloPhoenix.UserController :delete
 ```
 
-Scopes can also be nested, just like resources. Suppose that we had a versioned API with resources defined for images, reviews and users. We could then setup routes for the versioned API like this:
+Although technically scopes can also be nested (just like resources), the use of nested scopes is generally discouraged because it can sometimes make our code confusing and less clear. With that said, suppose that we had a versioned API with resources defined for images, reviews and users. Then technically we could  setup routes for the versioned API like this:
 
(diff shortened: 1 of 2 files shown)


C_routing.md
```diff
@@ -628,7 +628,7 @@ The `:browser` pipeline has five plugs: `plug :accepts, ["html"]` which defines
 
 Currently, the `:api` pipeline only defines `plug :accepts, ["json"]`.
 
-The router invokes a pipeline on a route defined within a scope. If no scope is defined, the router will invoke the pipeline on all the routes in the router. If we call `pipe_through/1` from within a nested scope, the router will invoke it on the inner scope only.
+The router invokes a pipeline on a route defined within a scope. If no scope is defined, the router will invoke the pipeline on all the routes in the router.
```
chrismccord: Let's add this: "If we call `pipe_through` within a nested scope, the router will invoke all `pipe_through`'s from parent scopes, followed by the nested one. "

chrismccord:
:ship: Thanks so much!

willykaram:
Thanks for the feedback. I made a change to that effect and can revise further based on any additional feedback. I made the revision in a separate commit, but once things are finalized I can condenses them with an interactive rebase/squash if helpful.

lancehalvorsen:
This looks great, @willykaram! Thanks so much. <3 <3 <3

Please squash the commits and we'll be ready to merge.

Issue #337, opened 2015-08-01
Channel guides should include token auth examples


chrismccord:
We have doc'd it here in phoenix, so should be a matter of transplanting:

https://github.com/phoenixframework/phoenix/blob/master/installer/templates/static/brunch/socket.js

willykaram:
@chrismccord @lancehalvorsen Not sure if the above PR transplants the example as intended, but it not please LMK and I can revised accordingly.

willykaram:
Apologies for the error with the prior PR/branch. I may have some further suggested revisions to this guide later in the week, but for now at least wanted to try and clean-up the prior commits and not leave the issue/PR hanging.

PR #454, opened 2016-01-02
Adding token auth example to Channels Guide per issue #337

willykaram:
Per issue #337, I've incorporated the token auth example into the Channel Guide based on what was in the Phoenix documentation at https://github.com/phoenixframework/phoenix/blob/master/installer/templates/static/brunch/socket.js

### DIFF
--- a/C_routing.md
+++ b/C_routing.md
@@ -498,7 +498,7 @@ review_path  DELETE  /reviews/:id       HelloPhoenix.ReviewController :delete
   user_path  DELETE  /users/:id         HelloPhoenix.UserController :delete
 ```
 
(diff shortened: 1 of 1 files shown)


G_channels.md
```diff
@@ -249,6 +249,72 @@ socket = assign(socket, :user, msg["user"])
...
+  end
+end
+```
+
+Step 3 - Pass Token on Connect:
+
+Finally, we need to pass the token on connect as below, or remove it from connect if we don't care about authentication.
+
+```javascript
+socket.connect()
```
chrismccord: we need to show passing the token in the `params` option of the Socket constructor. Otherwise this is set!

PR #456, opened 2016-01-03
Removed 'also' because 'as well' is used later.

parndt:
This is merely to remove a redundant word. :smile:

Thanks for the excellent docs.


lancehalvorsen:
Thanks for catching this and submitting a PR @parndt! Glad you're enjoying the guides. <3 <3 <3

PR #459, opened 2016-01-07
Fix typo in model testing guide

mcelaney:


lancehalvorsen:
Hi @mcelaney! Thanks for sending this.

I think the current version is correct because all of the values we set for the `count` variable are greater than one. If there were a possibility of the interpolated string reading, "should be at least 1 characters" or "should be at most 1 characters", then your PR would be correct. With the way the example code is written, though, that isn't possible.

Am I missing something?

mcelaney:
Hey @lancehalvorsen!
Sorry - I thought I had a comment on the PR... apparently I didn't hit 'comment.'

Basically - I ran in to problems following the tutorial as written. 

The min length test example copied from the tutorial:

    test "bio must be at least two characters long" do
      attrs = %{@valid_attrs | bio: "I"}
      assert {:bio, {"should be at least %{count} characters", [count: 2]}} in errors_on(%User{}, attrs)
    end
  
Gave me this:

    1) test bio must be at least two characters long (HelloPhoenix.UserTest)
       test/models/user_test.exs:24
       Assertion with in failed
       code: {:bio, {"should be at least %{count} characters", [count: 2]}} in errors_on(%User{}, attrs)
       lhs:  {:bio, {"should be at least %{count} characters", [count: 2]}}
       rhs:  [bio: {"should be at least %{count} character(s)", [count: 2]}]

After floundering around looking for a complicated reason the example didn't work I pinged the Phoenix Slack and @ventsislaf  pointed out the string issue.

Two commits showing this broken and then working here: https://github.com/mcelaney/hello_phoenix/commits/master

lancehalvorsen:
Ah, thanks @mcelaney, now I get it. The string changed in Ecto itself somewhere along the line.

[Ecto 1.1.1 source](https://github.com/elixir-lang/ecto/blob/3c5d052688194786f3532d7a0593e8c6177e1727/lib/ecto/changeset.ex#L1221-L1224)

Thanks again for putting this PR together! <3 <3 <3

PR #468, opened 2016-01-29
Add Bonus Guide on SEO Tags

danielberkompas:
Recommends that people use `render_existing` to customize their meta
tags.


jeregrine:
@lancehalvorsen will need some editorial here :-)

chrismccord:
@danielberkompas Thanks for the PR! I think we definitely want to cover use of `render_existing`, but I think it would be best served in a more generic guide on Phoenix Views that talked about customizing blocks of content per page, perhaps using SEO as an example. Once we refresh the View guide, we can look into including some of these ideas. Thanks!

PR #470, opened 2016-02-04
Use supervisor instead of worker for ecto repo

Linuus:
I believe the Ecto repo should use a supervisor and not a worker.


pinx:
This has changed recently in Phoenix. Maybe it's better to copy less of the source code of phoenix in the guides, so we don't have to keep things in sync. The user (developer) can look at the generated code herself.

jeregrine:
@pinx I agree, we need a slight re-work of the guides, in the mean time thanks @Linuus

PR #476, opened 2016-02-26
Add note about upgrading hex

markmeeus:
When following the guides, it wasn't immediately clear to me what I had to do when Hex was allready installed.
It turns out to be the same command, so I added a note in the guides.

kind regards,
Mark

### DIFF
--- a/introduction/E_installation.md
+++ b/introduction/E_installation.md
@@ -8,7 +8,7 @@ Phoenix is written in Elixir, and our application code will also be written in E
 
 If we have just installed Elixir for the first time, we will need to install the Hex package manager as well. Hex is necessary to get a Phoenix app running (by installing dependencies) and to install any extra dependencies we might need along the way.
 
-Here's the command to install Hex (If you have Hex allready installed, it will upgrade Hex to the lastest version):
+Here's the command to install Hex (If you have Hex already installed, it will upgrade Hex to the lastest version):
 
 ```console
 $ mix local.hex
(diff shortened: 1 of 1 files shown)


introduction/E_installation.md
```diff
@@ -8,7 +8,7 @@ Phoenix is written in Elixir, and our application code will also be written in E
 
 If we have just installed Elixir for the first time, we will need to install the Hex package manager as well. Hex is necessary to get a Phoenix app running (by installing dependencies) and to install any extra dependencies we might need along the way.
 
-Here's the command to install Hex:
+Here's the command to install Hex (If you have Hex allready installed, it will upgrade Hex to the lastest version):
```
stevedomin: s/allready/already

Gazler:
Just a tip for the future. If you change your commits and force push then the pull request will be updated.

For example in this case you could do:

`git rebase -i HEAD~2`

This would look like:

```
pick 6f891a7 Add note about upgrading hex
pick f8f9377 Fix typo

# Rebase 6f891a7..f8f9377 onto 6f891a7 (2 command(s))
```

You can then chose a command for each commit, so in this cause, you want to squash "Fix typo" into "Add note about upgrading hex". So you can change the `pick` to `s` (or `squash`):

```
pick 6f891a7 Add note about upgrading hex
s f8f9377 Fix typo

# Rebase 6f891a7..f8f9377 onto 6f891a7 (2 command(s))
```

After saving in your editor, you will be prompted with both commit messages, simply keep the first message and delete the "Fix typo" message as it is no longer relevant, it is like it didn't happen.

After saving the commit message, you should have a single commit with the changes from both commits. You can then push this up, however you will need to force push (normally you shouldn't force push to `master`, however that is the branch used for this pull request so I am using it here) as the history has changed:

`git push origin master -f`

This will update the pull request with your changes, so that you don't need to open a new one.

markmeeus:
Gary, 
Thank you for the tip.
I closed this pull request, and created a new one from a new branch with the 2 commits squashed together.

PR #478, opened 2016-02-27
Add note about upgrading hex

markmeeus:
When following the guides, it wasn't immediately clear how to upgrade Hex when it was allready installed.
It turns out to be the same command, so I added a note in the guides.

kind regards,
Mark


jeregrine:
Thanks @markmeeus

PR #479, opened 2016-02-27
Update console outputs

markmeeus:
The console outputs on the getting started page are a bit outdated.
It's not a big deal, but still, it could cause some confusion.
I updated them with the current outputs.

kind regards
Mark


jeregrine:
Thanks @markmeeus ! 💕

PR #480, opened 2016-02-28
updated deps to use mariaex ~> 0.6.1

domnikl:
To workaround an issue with MySQL where `mix ecto.migrate` will run infinitely without feedback. This is very frustrating for beginners. The error is already described here: https://github.com/elixir-lang/ecto/issues/1130. I think this will help a lot of people starting with Phoenix and MySQL.


jeregrine:
Thanks @domnikl 💖

PR #483, opened 2016-03-07
Fix mention of HTTP header for put_resp_content_type

christopheradams:
The guide mentions "HTTP Accepts Header" when it should be "Content-Type"


jeregrine:
Thanks @christopheradams :heart_eyes:

PR #484, opened 2016-03-09
Use to_string(k) for put_resp_header

AaronLenoir:
I was following the guide and the example for "put_headers" produced the error: no function clause matching in Plug.Conn.put_resp_header/3.

I figured this was because the key of the map was an atom, and put_resp_header expects a binary. So I converted to atom to_string and it worked.

If there's a better way to do this, I look forward to hear about that.

- Elixir 1.2.3 on Windows


jeregrine:
Thanks @AaronLenoir :heart:

PR #490, opened 2016-04-08
add ssl to heroku database config

bgentry:
Heroku Postgres requires SSL when connecting to databases:
https://devcenter.heroku.com/articles/heroku-postgresql#heroku-postgres-ssl

This updates the Heroku deployment guide to instruct users to enable SSL on their Ecto Postgres setup.

### DIFF
--- a/deployment/E_heroku.md
+++ b/deployment/E_heroku.md
@@ -105,7 +105,8 @@ Then, we'll add the production database configuration to `config/prod.exs`:
 config :hello_phoenix, HelloPhoenix.Repo,
   adapter: Ecto.Adapters.Postgres,
   url: System.get_env("DATABASE_URL"),
-  pool_size: 20
+  pool_size: 20,
+  ssl: true
 ```
 
 Now, let's tell Phoenix to use our Heroku URL and enforce we only use the SSL version of the website. Find the url line:
@@ -148,7 +149,8 @@ config :logger, level: :info
 config :hello_phoenix, HelloPhoenix.Repo,
   adapter: Ecto.Adapters.Postgres,
   url: System.get_env("DATABASE_URL"),
-  pool_size: 20
+  pool_size: 20,
+  ssl: true
 ```
 
 ## Creating Environment Variables in Heroku


ericmj:
Not sure why Heroku has this in the documentation because SSL is not required.

bgentry:
I wonder if it's only on non-free plans? I couldn't connect to my $50/mo
Heroku Postgres database without it, and I can say for sure that this
change worked for me.

ericmj:
I also use the $50 plan and it works for me without SSL.

https://github.com/hexpm/hex_web/blob/master/config/prod.exs#L10-L13

bgentry:
Weird. I was definitely seeing pg_hba.conf errors while trying to `heroku run mix ecto.migrate` until I made this tweak. @hgmnz do you happen to know the rules around when TLS is required for Heroku Postgres? I came across conflicting advice on Dev Center. Article I linked above said it's always required, while another said that was only true if connecting from outside AWS (which I wasn't doing).

Separately, there's the issue of whether this should be the recommended setting. If it works in all cases on Heroku, I think it definitely should be. Can you possibly check if this change also works for you?

ericmj:
I think we should be able to assume that the Heroku network is safe. Since SSL definitely adds some overhead I don't think it should be a default. The option is still there for those that want to use it.

lancehalvorsen:
@bgentry thanks for this contribution! I agree with @ericmj on not making SSL the default. Is there something you'd like to add instead? Perhaps a short sentence or two mentioning that it's possible to configure SSL for Postgres?

bgentry:
So I got more info on this tonight from my former Heroku colleagues. Turns
out that the specific plan I'm on ($50/mo standard-0) runs outside of EC2
classic in a VPC and therefore requires SSL to connect to it. This may be
true only for newer instances.

In the not too distant future it's likely that more if not all new Heroku
Postgres databases will require SSL. Since they all work with SSL turned
on, even old ones, it makes a lot of sense to document that approach for
newcomers.

ericmj:
@bgentry In that case please document that SSL may be needed and show how to add it to the configuration, thanks for investigating?

OT: So does that mean Heroku Postgres runs outside of AWS?

bgentry:
It's all on AWS, but some configs run outside EC2 Classic. EC2 Classic is
basically deprecated from Amazon's perspective in favor of VPCs and they
want all customers (including Heroku) to move away from it. Even when you
create a new AWS account your default environment is actually just a VPC
that's been preconfigured for you.

I believe Heroku dynos may also run in their own VPCs for a given fleet.

And on separate VPCs, traffic can't travel over 10.* IPs and therefore
appears to be not on the internal network.

I will make some changes to this PR today. Though I really do believe that
it's best to just have one set of instructions that works for everybody (if
possible).

hgmnz:
>  @hgmnz do you happen to know the rules around when TLS is required for Heroku Postgres? 

We are indeed migrating to requiring SSL across the board in all networks, so safest long term approach is to enable SSL in all cases. Currently only new DBs come up with SSL required, as we roll out this change.

bgentry:
Thanks for clarifying, @hgmnz! FYI, there were definitely some inconsistencies in Dev Center around this subject when I searched for "postgres SSL" or something similar.

In light of this info, I do believe that this PR should be merged as-is because:

1. TLS works for everybody using Heroku Postgres
2. In the not-too-distant future, TLS will be required for all Heroku Postgres instances
3. In most other languages, TLS gets used anyway with Heroku Postgres because libpq defaults to the `prefer` setting (using encryption if the server allows it). Seems like Ecto does not have an equivalent, instead having only `true` or `false` settings. I think there's also a good case to be made that the defaults here should be similar to what most people coming from other languages would expect with Postgres, and enabling encryption is the correct choice given that.

Please do let me know if there are other issues with the PR that you'd like me to resolve! :v:

ericmj:
> In light of this info, I do believe that this PR should be merged as-is because

Agreed. :shipit:

lancehalvorsen:
I'd like to thank you for clarifying as well @hgmnz. Thanks again for your contribution @bgentry! 🎉 <3

PR #491, opened 2016-04-09
Note to decrease ws timeout for Heroku deployment

keathley:
We were seeing H15 errors on a production application. Based on a conversation with @chrismccord we figured out that we simply needed to lower the websocket timeout window to less then Heroku's 55 second window. In this case that was 45 seconds. Since this was such an easy fix I thought that the information might be helpful to others. Let me know what you think.


chrismccord:
:shipit: :

lancehalvorsen:
Thanks so much for this @keathley! Much appreciated. <3

PR #493, opened 2016-04-18
Fix unclear wording in test guideline

Ingelheim:
I think matches is not clear enough and contains is more understandable in this case.


lancehalvorsen:
Thanks for this contribution @Ingelheim! I agree. 😄  <3

PR #514, opened 2016-05-26
Expand content on POOL_SIZE in Heroku guide

endersstocker:
Performed small copy edits and used console example for setting POOL_SIZE environment variable for mix task.

### DIFF
--- a/deployment/E_heroku.md
+++ b/deployment/E_heroku.md
@@ -116,8 +116,8 @@ If you do not know how many connections your Heroku Postgres instance provides, 
 
 When running a mix task you will also want to do:
 
-```sh
-POOL_SIZE=2 mix hello_phoenix.task
+```console
+$ heroku run "POOL_SIZE=2 mix hello_phoenix.task"
 ```
 
 So that Ecto does not attempt to open more than the available connections.
--- a/deployment/E_heroku.md
+++ b/deployment/E_heroku.md
@@ -68,7 +68,7 @@ A [buildpack](https://devcenter.heroku.com/articles/buildpacks) is a convenient 
 
 The URL in the output is the URL to our application. If we open it in our browser now, we will get the default Heroku welcome page.
 
-> Note: if we hadn't initialized our Git repository before we ran the `heroku create` command, we won't have our Heroku remote repository properly set up at this point. We can set that up manually by running: `heroku git:remote -a [our-app-name].`
+> Note: if we hadn't initialized our Git repository before we ran the `heroku create` command, we wouldn't have our Heroku remote repository properly set up at this point. We can set that up manually by running: `heroku git:remote -a [our-app-name].`
 
 ## Adding the Phoenix Static Buildpack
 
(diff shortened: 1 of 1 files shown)


deployment/E_heroku.md
```diff
@@ -215,7 +215,7 @@ Setting config vars and restarting mysterious-meadow-6277... done, v3
 SECRET_KEY_BASE: xvafzY4y01jYuzLm3ecJqo008dVnU3CN4f+MamNd1Zue4pXvfvUjbiXT8akaIF53
 ```
 
-If you need to make any of your config variables available at compile time you will need to explicitly define which ones in a configuration file. Create a file `elixir_buildpack.config` in your applications root directory and add a line like: `config_vars_to_export=(DATABASE_URL MY_VAR)`. See [here](https://github.com/HashNuke/heroku-buildpack-elixir#specifying-config-vars-to-export-at-compile-time) for more information. 
+If you need to make any of your config variables available at compile time you will need to explicitly define which ones in a configuration file. Create a file `elixir_buildpack.config` in your applications root directory and add a line like: `config_vars_to_export=(DATABASE_URL MY_VAR)`. See [here](https://github.com/HashNuke/heroku-buildpack-elixir#specifying-config-vars-to-export-at-compile-time) for more information.
```
jeregrine: Could you add `POOL_SIZE` as a default here too. It wasn't part of the original but now that I see it, we're going to need it.

jeregrine:
@endersstocker thank you so much!

deployment/E_heroku.md
endersstocker: @jeregrine Good call. See cb7db7e.

endersstocker: @jeregrine Good call. See 9d3016a.

endersstocker: `POOL_SIZE` should only be needed at runtime, not compile time.

deployment/E_heroku.md
```diff
@@ -163,7 +149,7 @@ config :logger, level: :info
 config :hello_phoenix, HelloPhoenix.Repo,
   adapter: Ecto.Adapters.Postgres,
   url: System.get_env("DATABASE_URL"),
-  pool_size: 20,
+  pool_size: String.to_integer(System.get_env("POOL_SIZE") || "10"),
```
jeregrine: I believe this variable will need to be available at build time so setting it here https://github.com/phoenixframework/phoenix_guides/pull/514/files#diff-969f42064dcef84e92928f7483e56814R220 will be required.

endersstocker: @jeregrine I tested it with and without `config_vars_to_export` in the buildpack configs. It seems that `POOL_SIZE` is only needed at compile time if `|| "10"` is left out. I verified that the Phoenix app actually uses the value of `POOL_SIZE` and not `"10"` in absence of `config_vars_to_export`.

jeregrine: Thanks!

jeregrine:
Thank you so much!

PR #515, opened 2016-05-26
Corrected description of `cast/3` to `cast/4`.

sillypog:
Small change to correct the description of the arity of the Ecto.Changeset cast function. This commit also removes some unnecessary whitespacing from an unrelated line in the same file.


jeregrine:
Thanks!

PR #519, opened 2016-05-30
Update Exrm Releases links

cjwebb:
Two links in the `advanced deployment` guide currently 404 when clicked on. Replace with working URLs.


jeregrine:
Thanks!

PR #521, opened 2016-05-31
Explain composite primary key

pinx:
Ecto supports composite primary keys, so we need to explain how to use them.


bonus_guides/EE_custom_primary_key.md
```diff
@@ -117,3 +117,37 @@ Indexes:
...
+defmodule HelloPhoenix.Player do
+  use HelloPhoenix.Web, :model
+
+  @primary_key false
+  schema "players" do
+    field :first_name, primary_key: true
+    field :last_name, primary_key: true
+    field :position, :string
+    field :number, :integer
+  . . .
```
jeregrine: Could you show an example of querying for the composite key as well? Thanks!

jeregrine:
@pinx thanks!

PR #522, opened 2016-06-02
Modified outdated jquery source url

ayamamori:
Hi.

I'm very beginner of phoenix and I found some outdated sample source code in the "channels tutorial", while I followed it.
The url ``code.jquery.com/jquery-1.11.9.min.js`` is now 404 response and should be modified to new one.


jeregrine:
Thank you!

PR #533, opened 2016-07-01
Update DD_Plug.md

Ingelheim:
Hi there,

while working on some i18n in my Phoenix app, I think I found a mistake in the documentation. The 'conn' variable is not known in the call method. I don't know what the right way would be, but wanted to bring your attention to it.

Best
Lukas


Gazler:
I think the existing code is correct. Could you please show the error you are getting. The first argument to the `call/2` function should be the Plug.Conn struct.

If you still think the existing code is incorrect, could you please open a new issue?

PR #535, opened 2016-07-06
Update G_static_assets.md

asifshiraz:
Included explanation of whitelisting resources for serving over http from the web/statics/assets directory

### DIFF
--- a/bonus_guides/G_static_assets.md
+++ b/bonus_guides/G_static_assets.md
@@ -20,7 +20,7 @@ The second important file is `brunch-config.js`. This is configuration for Brunc
 
 According to this configuration Brunch will look for asset source files in `web/static`.
 
-Files and directories in `web/static/assets` will be copied to the destination without changes.
+Files and directories in `web/static/assets` will be copied to the destination without changes. Note that while all directories will be copied, only those will be served over http, which have been whitelisted in Plug.Static configuration, which by default are only the js, css, images and fonts directories.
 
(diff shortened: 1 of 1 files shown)


bonus_guides/G_static_assets.md
```diff
@@ -20,7 +20,7 @@ The second important file is `brunch-config.js`. This is configuration for Brunc
 
 According to this configuration Brunch will look for asset source files in `web/static`.
 
-Files and directories in `web/static/assets` will be copied to the destination without changes.
+Files and directories in `web/static/assets` will be copied to the destination without changes. Note that while all directories will be copied, only those will be served over http, which have been whitelisted in Plug.Static configuration, which by default are only the js, css, images and fonts directories.
```
Gazler: Could we reword this entire block as something like:

> All of the files in `web/static/assets` will be copied to the destination without changes.
>
> Phoenix by default will serve files in the `css`, `fonts`, `images` and `js` directories. The directories and files served by Phoenix can be set by changing the `only` option of the `Plug.Static` plug in `lib/hello_phoenix/endpoint.ex`

 > The default configuration is:

```elixir
  plug Plug.Static,
    at: "/", from: :hello_phoenix, gzip: false,
    only: ~w(css fonts images js favicon.ico robots.txt)
```

jeregrine:
@asifshiraz I agree with @Gazler's suggestion, that change does a double whammy of helping the user learn where to look for these things.

PR #539, opened 2016-07-13
Update from rooms to room

skippednote:
`mix phoenix.new` generates a `user_socket.ex` with `room` not `rooms`.

```ex
  ## Channels
  # channel "room:*", HelloPhoenix.RoomChannel

```


Gazler:
Thanks <3

PR #541, opened 2016-07-15
clarify url generation behind a reverse proxy

schnittchen:


bonus_guides/S_phoenix_behind_proxy.md
```diff
@@ -24,12 +23,28 @@ use Mix.Config
 
 config :hello_phoenix, HelloPhoenix.Endpoint,
   http: [port: {:system, "PORT"}],
-  url: [host: "example.com"],
   cache_static_manifest: "priv/static/manifest.json"
 
 . . .
 ```
 
+Urls generated using a `_url` function from the `HelloPhoenix.Router.Helpers` module will include `localhost:port_as_above`. To fix that, you need to add the following `url` option:
```
Gazler: Could we rewrite this as:

 > Urls generated using a `_url` function from the `HelloPhoenix.Router.Helpers` module will include a url such as http://localhost:8080/users for `user_url(conn, :index)`. To fix this we can use the `url` option:

bonus_guides/S_phoenix_behind_proxy.md
```diff
@@ -24,12 +23,28 @@ use Mix.Config
...
+
+config :hello_phoenix, HelloPhoenix.Endpoint,
+  http: [port: 8080],
+  url: [host: "example.com", port: 80],
+  cache_static_manifest: "priv/static/manifest.json"
+
+. . .
+```
+
+The port value is necessary. With port 80 as above, generated links will not include the port explicitly.
```
Gazler: Can we change this to:

> Our url will now be http://example.com/users for the `user_url(conn, :index)` function. Not that the port is not present in the url. If the scheme is `http` and the port is `80`, or the scheme is `https` and the port is `443` then the port will not be present in the url. In all other circumstances it will be present.

bonus_guides/S_phoenix_behind_proxy.md
schnittchen: I like that.

bonus_guides/S_phoenix_behind_proxy.md
schnittchen: Yes!

schnittchen:
Added commit with @Gazler 's suggestions

Gazler:
Thanks! <3

PR #545, opened 2016-08-02
Add note about Mix.env when sending emails

DamirSvrtan:


Gazler:
This is recommended by the library here: https://github.com/chrismccord/mailgun/blob/ace116f37a86b3b11cecad6f37c4599f4552452b/lib/client.ex#L16

Gazler:
Thanks <3

PR #549, opened 2016-08-08
update email guide: Make it clear that the mailgun package is not the

lau:
update email guide: Make it clear that the mailgun package is not the only option, not part of Phoenix, and link to some alternatives.

On twitter I saw a new user that was confused and thought that Mailgun was the only option for sending email with Phoenix.


Gazler:
Thanks <3

PR #553, opened 2016-08-10
Grammar pedantry

PaulOstazeski:
Adding "-ing" to the verb "follow" so that the progressive tense is used.


Gazler:
Thanks. Looks like this was fixed in https://github.com/phoenixframework/phoenix_guides/pull/544

PaulOstazeski:
Cool

PR #534, opened 2016-07-01
Model Guide Updates (Phoenix 1.2)

ybur-yug:
# Model Guide Updates (Phoenix 1.2)
In relation to #532 I had some time this Friday to dive into this for work. I plan on getting through the whole model guide today, and just figured I'd throw this up for discussion as it goes on.

## Process
I took a pretty simple process to go through.

First I just generated a new project after making sure I had Phoenix 1.2/Elixir 1.3 installed. From here, I followed step by step and just updated all code and changes that I saw. There were several issues open that were essentially the changes needed to this guide as a blanket, as well as a few others I saw.

## Changes
Commit message:
```
Update for Phoenix 1.2 in ecto model guides
This includes issues 529, 530, and 531

Issues Handled:
Issue 529
This brought up the initial concerns about the various pieces that were
later broken into smaller issues after a suggestion by @jeregrine

Issue 530
These bits were specifically tied to the generators. They referenced the
now-unused `@required_fields` and `@optional_fields` rather than 1.2's
new style of declaration for this. All code references were changed by
using a project following the guide.

Issue 531
This removes the reference to the no-longer-generated line and also
corrects the statements around it to still be accurate. As there is
already plans mentioned for adding a section on scrubbing/securing
parameters properly, reference is most likely best left there in the
future.

The whole of the changes boil down to code references, and correction of
explanation of various functions used in the model pipeline. Controllers
and their portions as well as the installation-as-standalone pieces had
minimal modification.
```

## Issues referenced:

- [x] [529](https://github.com/phoenixframework/phoenix_guides/issues/529)
- [x] [530](https://github.com/phoenixframework/phoenix_guides/issues/530)
- [x] [531](https://github.com/phoenixframework/phoenix_guides/issues/531)

PR #556, opened 2016-08-13
Update models testing to match changes

kimroen:
A few things I came across when going through this guide. Very related to the changes in https://github.com/phoenixframework/phoenix_guides/pull/534.

- `cast/4` is deprecated in favor of `cast/3` + `validate_required/3`
- the first parameter to `changeset/2` is `struct`, not `model`
- the fallback last param for `changeset/2` has changed


jeregrine:
Great! Thank you @kimroen

Issue #442, opened 2015-12-15
Potential confusion when describing function plugs


dmarkow:
In the [Function Plug](https://github.com/phoenixframework/phoenix_guides/blob/master/DD_Plug.md#function-plugs) section there are some potentially confusing code blocks:

- The first code block for `show/2` refers to several functions (`authenticate/1`, `find_message/1`, etc.) but doesn't actually show them. It should maybe either have a `...` like the code block above it to at least imply that there is more stuff in this module after `show/2`? Or maybe explicitly list out the code for those functions?
-  Similarly, the second example for `show/2` is restructured into the multiple plugs which refer to those unlisted functions. For example, the new `find_message/2` function calls the original (unlisted) `find_message/1` function. This could be confusing to someone new to Phoenix (and Elixir), and explicitly showing the code for both `find_message/1` and `find_message/2` could be helpful.

I will happily make a pull request for this but wanted to get some input on how verbose these code blocks should/shouldn't be first. Thanks!

henrik:
@dmarkow Maybe showing the method definitions with elided bodies would be good? Like this:

``` elixir
defp authenticate(conn), do: …
defp find_message(id), do: …
defp authorize_message(conn, id), do: …
```

Doesn't take up a lot of space or add a lot of complexity but makes things less confusing-

huyz:
Funny, I saw the `...` and wondered if this was just a documentation artifact or if it had special meaning in Phoenix/Elixir.   Someone on Slack pointed me to this page.  It would have been clearer if the guide mentioned that `...` means that some code was omitted for the sake of brevity.

henrik:
@huyz Can totally see how one would think that. Maybe we could replace `, do: ...` with something syntactically valid like `, do: "…implementation…"`?

PR #557, opened 2016-08-15
Add authenticate/1 and find_message/1 to Plug doc. (#442)

jsncmgs1:
This issue has been open for a while, https://github.com/phoenixframework/phoenix_guides/issues/442

I haven't seen anyone disagree with the suggestions in the comments, so I added those in.  If we think something better could be in place, I'm happy to change it.


jeregrine:
Thanks @jsncmgs1 !

PR #559, opened 2016-08-16
Update channel testing guide

atsheehan:
When running `mix phoenix.gen.channel Room rooms`, the generator
fails with the following error message:

```
** (Mix) mix phoenix.gen.channel expects just the module name:

    mix phoenix.gen.channel Room
```

This commit updates the channel testing guide to use the source
generated from running the `mix phoenix.gen.channel Room` command.

Also, the `Testing an Asynchronous Push from the Server` section 
references a `handle_out/3` function in the `RoomChannel` which does 
not exist in the generated code. I did not change this section, I wasn't sure
what the appropriate change was.

Includes @jordan0day's changes from #550.

### DIFF
--- a/testing/J_channels.md
+++ b/testing/J_channels.md
@@ -24,14 +24,12 @@ $ mix phoenix.gen.channel Room rooms
 * creating web/channels/room_channel.ex
 * creating test/channels/room_channel_test.exs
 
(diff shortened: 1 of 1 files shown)


jeregrine:
Great Thank you!

PR #561, opened 2016-08-19
Remove root option as it's not compulsory from now

noma4i:
https://github.com/phoenixframework/phoenix/pull/1838
`. root: "." is no longer required since Phoenix v1.2.`

Not sure about `Mix.Project.config[:version]`. It's missing in https://github.com/phoenixframework/phoenix/blob/master/installer/templates/new/config/prod.exs


Gazler:
Thanks <3

PR #563, opened 2016-08-20
Use "page.json" and "author.json" in View JSON

jsncmgs1:
In the JSON view doc's index should render_many "author.json" and "page.json" instead of render_many "show.json".


Gazler:
Thanks <3

Issue #99, opened 2014-12-23
Testing Guide


lancehalvorsen:
There have been several questions on the mailing list lately about how to test various parts of a Phoenix application. It would be great to have a guide as a reference.

jeregrine:
Since @josevalim did such a nice job with ConnTest we should be able to tackle this one.

josevalim:
Thank you! Channel test should come this weekend. We also have ModelCase on
v0.12.0z

-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Lead Developer*

lancehalvorsen:
:+1: to the delights of ConnTest and ModelCase!

Also, since the generators now provide test files, getting folks up to speed on testing models and controllers will be significantly easier, I think.

Looking forward to having these conveniences on Channels as well.

lancehalvorsen:
I just opened up PR #238 as a skeleton for this.

I'm working on the intro now, and will head into the models and controllers in turn. The larger plan is to use the the HTML resource we built for ecto models as a jumping off place because a) folks should be familiar with the code already, and b) there are already tests in place that we can add on to. :^)

gjaldon:
Would you like some help here, @lancehalvorsen? I've had the chance to use `Phoenix.ChannelTest` when working with the channel generator in Phoenix and have used ConnTest and ModelCase in an app I'm building.

lancehalvorsen:
Yes, please! Thanks so much for the offer, @gjaldon.  :)

lancehalvorsen:
@gjaldon, I just closed this https://github.com/phoenixframework/phoenix_guides/pull/152, but there may still be things you can use from it. @paulcsmith has kindly offered it up if it's useful.

paulcsmith:
@gjaldon LMK if you need any help with the channel guide. As @lancehalvorsen mentioned, the one I started might be a good outline for the new one

gjaldon:
@paulcsmith thanks! Will let you know if I get stuck and will likely use the one you started as outline.

jsncmgs1:
I'm going to start on a guide for Controller specs, specifically testing JSON responses.  I personally never write controller tests if I'm rendering templates (I write feature tests), so if someone else would like to tackle that chunk that would be great.  If not, I'll do it after I'm done with the JSON stuff.

Or perhaps the guide for testing non API controllers is a guide on features tests?

jsncmgs1:
[...]

PR #566, opened 2016-08-20
Add Controller Testing Guide (#99)

jsncmgs1:
This covers testing an API controller, there is nothing in here
for feature test style tests.

Makes progess on https://github.com/phoenixframework/phoenix_guides/issues/99 , but doesn't close.

### DIFF
--- a/testing/G_controllers.md
+++ b/testing/G_controllers.md
@@ -1,1 +1,296 @@
-## FPO
+# Testing Controllers
+
+We're going to take a look at how we might test drive a controller which has endpoints for a JSON api.
+
+Phoenix has a generator for creating a JSON resource which looks like this:
+
+### Set up
+```bash
+mix phoenix.gen.json Thing things some_attr:string another_attr:string
+```
+
+Thing is the Model, things is the table name, and some_attr and another_attr are database columns on table things of type string. Don't run this command, we're going to explore test driving out a similar result to what a generator would give us.
+
+Let's create a `User` model. Since model creation is not in scope of this guide, we will use the generator.  If you aren't familiar, read [this section of the Mix guide](http://www.phoenixframework.org/docs/mix-tasks#section--mix-phoenix-gen-model-).
+
+```bash
+$ mix phoenix.gen.model User users name:string email:string
+```
+
+Now run migrations:
+
+```bash
+$ mix ecto.migrate
+```
+
+### Test driving
+
+What we are going for is a controller with the standard CRUD actions. We'll start with our test since we're TDDing this. Create a `user_controller_test.exs` file in `test/controllers`
+
+```elixir
+# test/controllers/user_controller_test.exs
+
+defmodule HelloPhoenix.UserControllerTest do
+  use HelloPhoenix.ConnCase, async: true
+
+end
+```
+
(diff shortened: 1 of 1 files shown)


testing/G_controllers.md
```diff
@@ -1 +1,297 @@
...
+    test "Edits, and responds with a the user if attributes are valid"
+    test "Returns an error and does not edit the user if attributes are invalid"
+  end
+
+  test "delete/2 and responds with :ok if the user was deleted"
+
+end
+```
+
+Here we have tests around the 5 controller CRUID actions we need to implement for a typical JSON API. In 2 cases, index and delete, we are only testing the happy path, because in our case they generally won't fail because of domain rules (or lack thereof). In practical application, our delete could fail easily once we have associated resources that cannot leave orphaned resources behind, or number of other situations. On index, we could have filtering and searching to test. Also, both could require authorization.
```
Gazler: Typo: "CRUID" -> "CRUD"

jsncmgs1:
Thanks! Just fixed it and squashed.

testing/G_controllers.md
```diff
@@ -1 +1,297 @@
...
+      "data" => [
+        %{ "name" => "John", "email" => "john@example.com" },
+        %{ "name" => "Jane", "email" => "jane@example.com" }
+      ]
+    }
+
+    assert response == expected
+  end
+```
+Let's take a look at once going on here. Each test gets a map with a test conn passed into it, so we pattern much on that to get our conn.  We build our users, and use the `get` function to make a `GET` request to our `UserController` index action, which is piped into `json_response/2` along with the expected HTTP status code.  This will return the JSON from the response body, when everything is wired up properly. We represent the JSON we want the controller action to return with the variable `expected`, and assert that the `response` and `expected` are the same.
```
sbatson5: Small typo: `...look at what's going on...` rather than `once going on`.

testing/G_controllers.md
```diff
@@ -1 +1,297 @@
...
+end
+```
+
+When we run the test again, it passes.
+
+The last item we'll cover is the case where we don't find a user in `show/2`.
+
+Try this one on your own and see what you come up with.  One possible solution will be given below.
+
+Walking through our TDD steps, we add a test supplies a non existent user id to `user_path` which returns a 404 status and an error message:
```
sbatson5: Think we are missing a word: `we add a test supplies a non existent...` should be `we add a test that supplies a non...`

jsncmgs1: Thanks! Fixed.

Gazler:
Thanks <3

testing/G_controllers.md
```diff
@@ -1 +1,296 @@
...
+      "data" => [
+        %{ "name" => "John", "email" => "john@example.com" },
+        %{ "name" => "Jane", "email" => "jane@example.com" }
+      ]
+    }
+
+    assert response == expected
+  end
+```
+Let's take a look at what's going on here. Each test gets a map with a test conn passed into it, so we pattern much on that to get our conn.  We build our users, and use the `get` function to make a `GET` request to our `UserController` index action, which is piped into `json_response/2` along with the expected HTTP status code.  This will return the JSON from the response body, when everything is wired up properly. We represent the JSON we want the controller action to return with the variable `expected`, and assert that the `response` and `expected` are the same.
```
mydearxym: seems `pattern much`  should be `pattern match`

PR #571, opened 2016-08-26
Use npm run deploy instead of brunch build

Valentin-Seehausen:
If someone won't have a local installation of brunch, the brunch build command will fail. So just let npm handle the stuff. :)

### DIFF
--- a/deployment/A_introduction.md
+++ b/deployment/A_introduction.md
@@ -59,12 +59,12 @@ $ MIX_ENV=prod mix compile
 
 ## Compiling your application assets
 
-This step is required only if you have static assets like images, javascript, stylesheets and more in your Phoenix applications. By default, Phoenix uses brunch, and that's what we are going to explore.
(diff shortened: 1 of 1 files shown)


deployment/A_introduction.md
```diff
@@ -59,12 +59,12 @@ $ MIX_ENV=prod mix compile
 
 ## Compiling your application assets
 
-This step is required only if you have static assets like images, javascript, stylesheets and more in your Phoenix applications. By default, Phoenix uses brunch, and that's what we are going to explore.
+This step is required only if you have static assets like images, javascript, stylesheets and more in your Phoenix applications. By default, Phoenix uses brunch, and that's what we are going to explore. You don't need a local installation of brunch, just run ```npm run deploy``` and it will invoke ```brunch build --production``` for you.
```
Gazler: I think the "You don't need a local installation of brunch" is a little misleading, it still needs to be installed.

PR #576, opened 2016-09-02
Remove redundant sentence from ecto-models docs.

tommcgurl:
Came across this small redundancy in the docs, so simply removing the first occurrence. 👍


Gazler:
Thanks <3

Issue #462, opened 2016-01-20
Migrations & Related Coverage in Guides


willykaram:
Ensure that the Guides provide sufficient coverage of Migrations, including CLI commands, table/column mods, and pointers to Ecto docs.

Per the discussion on this HN thread: 

https://news.ycombinator.com/item?id=10931469

chrismccord:
:+1: thanks for opening this from the HN discussion

ralphking:
Hi, this was based on my feedback so I'll try to help out here. The issue I had was not being able to quickly find out how to create a migration for something that already exists. I was probably (definitely) having a bad day, so mostly my fault.

The actual guide for that is here http://www.phoenixframework.org/docs/mix-tasks, and is a great guide. However as it was under the 'bonus guides' section, I didn't immediately see it, and ended up looking at a blog post from a while back just to find the right mix task.

Perhaps there could be a note here http://www.phoenixframework.org/docs/ecto-models somewhere, pointing to that bonus guide. i.e 'for a more comprehensive list of mix tasks see here'.

As it's a fairly important part of the framework perhaps it should be there rather than in a 'bonus guide'.

Anyway that's just my 2p.

willykaram:
@ralphking thanks a lot for adding the feedback here & on HN. The Guides especially need to work on 'bad days', so I'll make some revisions as you've suggested to help make things more clear for people like you and me, and hopefully we can get a PR merged to resolve the issue ... thx again.

jeregrine:
@ralphking @willykaram Any PR's happily accepted.

willykaram:
@jeregrine thx, I can take care of of a PR once I'm done the Seeding Data Guide (#465) and revising the Request Lifecycle Diagram (#118) to reflect what @scrogson covered in his Erlang Factory Talk

PR #578, opened 2016-09-06
Add link to Ecto section of mix tasks guide

philosodad:
This is in response to [issue #462](https://github.com/phoenixframework/phoenix_guides/issues/462). A more thorough change might be in order but this patches the issue by adding a link to the anchor tag of the mix tasks guide section for Ecto.


Gazler:
Thanks <3

PR #582, opened 2016-09-13
Fix code example for composite keys

faheemmughal:
Guide for custom primary key had invalid code example. The code should had field type as second argument.


Gazler:
Thanks <3

PR #583, opened 2016-09-18
fix minor misspellings in installation and heroku

sbatson5:
Was reading through the guides and just spotted a couple minor typos, so I thought I would address them 😄


Gazler:
Thanks <3

PR #584, opened 2016-09-22
Add missing steps in manual Ecto setup for existing applications

EyMaddis:
I noticed that the guide just adds the the dependencies to `deps`, but they are not listed in the application dependencies. 
This will result in startup errors which might discourage new users joining the ecosystem.

Furthermore it contains the changes for Ecto 3, which are also mentioned in the [migration guide](http://www.phoenixframework.org/blog/upgrading-from-11x-to-120).

### DIFF
--- a/H_ecto_models.md
+++ b/H_ecto_models.md
@@ -625,6 +625,15 @@ defmodule HelloPhoenix.Mixfile do
 end
 ```
 
+We also need to explicitly append both `:postgrex` and `:phoenix_ecto` in the list of applications.
+
+```elixir
+def application do
+    [mod: {HelloPhoenix, []},
+     applications: [:phoenix, :phoenix_pubsub, :phoenix_html, ... :postgrex, :phoenix_ecto]]
+end
+```
+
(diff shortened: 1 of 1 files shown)


EyMaddis:
I added that the `web/web.ex` should include model dependencies in order to generate models

jsncmgs1:
@EyMaddis you should squash [aeb5c84], possibly all 3 but definitely the typo :)

EyMaddis:
@jsncmgs1 there are now only two commits left :)

Gazler:
Thanks, I've just added one comment then this is good to go.

EyMaddis:
@Gazler I adapted your suggestion, but also included what the line does, so it is more beginner friendly. I did learned the semantics of this within the process of migration, so I think that it might be helpful to state that it will import everything globally.

Gazler:
Thanks <3

Closed via https://github.com/phoenixframework/phoenix_guides/commit/775a006df3e8da0b72eb437d912fb94b60ab53ef

PR #589, opened 2016-10-01
Fix brunch docs link on the static assets guide

carlosantoniodasilva:
The documentation was moved out of the main repo, so point the reference to the getting started guide on the website instead.

https://github.com/brunch/brunch/pull/1376


Gazler:
Thanks <3

PR #593, opened 2016-10-11
Fix deprecated conn, change to build_conn

hirocaster:
I found deprecated code in `this file` at run test.

`warning: using conn/0 to build a connection is deprecated. Use build_conn/0 instead.`

Fix it code.


Gazler:
Thanks <3

PR #595, opened 2016-10-13
Add simple mix.exs to allow previewing guides

wsmoak:
This is intended for the ex_doc branch.

Rather than duplicating the 'extras' config here, just pick up all the .md files and generate the html so that contributors can preview changes easily.

Also, ignore the generated /doc directory.

Note: this won't work until the .md files are moved under the docs/ directory (which I believe is the plan...)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store
 /_build
 /deps
+/doc
 erl_crash.dump
 .sass-cache
 *.ez
--- a/mix.exs
+++ b/mix.exs
@@ -1,0 +1,35 @@
+defmodule PhoenixGuides.Mixfile do
+  use Mix.Project
+
+  @version "1.3.0-dev"
+
+  def project do
+    [app: :phoenix_guides,
+     name: "Phoenix Guides",
+     version: @version,
+     elixir: "~> 1.3",
+     deps: deps(),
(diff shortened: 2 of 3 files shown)


Gazler:
@wsmoak I've done this, just wanted to make sure this is what everyone expected before I push it up as the ex_doc branch.

https://github.com/Gazler/phoenix_guides/commit/12e128196ef373f30c812d6ea1c89406525223e0

// @chrismccord @josevalim @jeregrine

wsmoak:
That looks great @Gazler, it's what I was expecting.

I think doing this 1:1 move/rename on the first pass is good.  There are a couple of duplicate filenames to deal with since ExDoc flattens the directory structure, but I can do it in the config and then suggest another (much smaller) pass for those.

wsmoak:
Sorry @Gazler I didn't check this closely enough... in the phoenix_guides/ex_doc branch, it seems we've lost the "deployment" subdirectory, and when those files were moved to the top level, overwritten the original introduction.md file.  Can you take a look?

Gazler:
Maybe we should remove the logo on this? When previewing the guides, the logo looks a little weird. We should either remove it, or use the same one from the Phoenix guides.

guides

chrismccord:
+1 to same one from phoenix docs

PR #596, opened 2016-10-13
Updated mix tasks documentation

javiercr:
Added missing mix tasks to the list and documented the behavior of `mix phoenix.gen.presence`.

### DIFF
--- a/bonus_guides/D_mix_tasks.md
+++ b/bonus_guides/D_mix_tasks.md
@@ -378,8 +378,8 @@ Add the channel to your `web/channels/user_socket.ex` handler, for example:
 
 #### `mix phoenix.gen.presence`
 
-This task will generate a Presence tracker for your application. You can pass the module name as an optional argument,
-otherwise it will default to `Presence`.
+This task will generate a Presence tracker. The module name can be passed as an argument, 
+`Presence` is used if no module name is passed.
 
 ```console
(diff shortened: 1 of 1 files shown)


javiercr:
I haven't included specific documentation for `mix local.phoenix` and `mix phoenix.gen.secret` as I think they're pretty simple and explained in the initial list.

bonus_guides/D_mix_tasks.md
```diff
@@ -373,6 +376,16 @@ Add the channel to your `web/channels/user_socket.ex` handler, for example:
     channel "rooms:lobby", HelloPhoenix.RoomChannel
 ```
 
+#### `mix phoenix.gen.presence`
+
+This task will generate a Presence tracker for your application. You can pass the module name as an optional argument,
```
Gazler: Could we change the narritive so that it reads like the rest of the descriptions?

Something like:

 > +This task will generate a Presence tracker. The module name can be passed as an argument, `Presence` is used if no module name is passed.

Gazler:
One small comment then this is good to merge.

javiercr:
Makes sense, done!

Gazler:
Thanks <3

Closed via https://github.com/phoenixframework/phoenix_guides/commit/b503e416e67763ebf002e6df30532141fc5b4024

PR #600, opened 2016-10-20
Add missing "the"

evuez:
In [Fault Tolerance and Reliability Guarantees](http://www.phoenixframework.org/docs/channels#section-fault-tolerance-and-reliability-guarantees), second bullet point (*Resending Client Messages*):

> Channel clients [...] send them to **the** server [...].


Gazler:
Thanks <3

PR #602, opened 2016-10-22
Avoid duplicate filenames for ExDoc

wsmoak:
Note: this is intended for the ex_doc branch, not master.

Because ExDoc flattens the directory structure, duplicate filenames can't be used unless they are re-named in the configuration.

For previewing the generated docs we're using a `docs/**/*.md` wildcard rather than listing out all the files.

This renames the duplicate files under `docs/testing/` so they don't collide.  (The models.md file wasn't a duplicate but was changed to follow the naming pattern.)


Gazler:
Fixed on ex_doc. Thanks for pointing it out! <3

PR #609, opened 2016-11-03
Update headers on Plug docs and fix links

wsmoak:
- Use h2's on Plug docs so the items will show in the left-hand menu
- Use relative links to plug.html instead of full urls


Gazler:
Thanks <3

PR #610, opened 2016-11-04
Updates deps version for "Sending Email" guide.

dummey:
Specifically the bamboo_smtp version that is used in the guide has dependency
issues when trying to install via `mix deps.get`. Using the latest versions still
works with the guide.


Gazler:
Thanks <3

PR #614, opened 2016-11-06
Use relative urls instead of absolute urls

Gazler:
All URLs that previously referenced http://phoenixframework.org for the
pages will no longer be valid when the guides are merged with the
Phoenix repository.

### DIFF
--- a/docs/adding_pages.md
+++ b/docs/adding_pages.md
@@ -126,7 +126,7 @@ end
 
 ```
 
-For now, we'll ignore the pipelines and the use of `scope` here and just focus on adding a route. (We cover these topics in the [Routing Guide](http://www.phoenixframework.org/docs/routing), if you're curious.)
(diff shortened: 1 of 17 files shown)


Gazler:
Once this is merged, we will no longer be able to copy whole pages to readme.io as the links will not work. Perhaps we should do a deployment freeze to readme.io? (Except for bug fixes/broken links)

wsmoak:
I think you already merged a PR where I changed one of these.  Can you branch from a revision before that, and publish from a branch to Readme.io?

That would allow important fixes to get published while letting us continue on master for the 1.3 docs.

Gazler:
@wsmoak You are right, I did merge one of yours, I didn't change the link on readme.io though, so everything is still working. Good idea on creating a branch and back merging in, I'll do that if there are too many changes to handle manually.

PR #619, opened 2016-11-11
added a supervision tree init line for using repo

cultofmetatron:
Its not unheard of to access the database from a mix task. However there's nothing on the site that indicates that anything additional needs to be added in order to access the database.

This should same the next guy any confusion.

### DIFF
--- a/docs/bonus_guides/mix_tasks.md
+++ b/docs/bonus_guides/mix_tasks.md
@@ -778,6 +778,7 @@ defmodule Mix.Tasks.HelloPhoenix.Greeting do
   """
 
   def run(_args) do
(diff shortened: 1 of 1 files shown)


Gazler:
I haven't seen this technique used before. Starting the whole application may result in parts of the application running when you don't need them to. This could result in `mix phoenix.server` running on port 4000, and the mix task also tries to start the server on the same port, resulting in a port conflict.

Usually I do something like the code snippet in: https://elixirforum.com/t/myapp-repo-configuration-not-found-in-an-exrm-release-migration-task/1268/6 to ensure the ecto dependencies are started. 

Perhaps we could have a subsection of "Using Ecto in mix tasks" and mention this approach there?

PR #620, opened 2016-11-11
Add mix docs.watch task for watching markdown files

Gazler:
`mix docs.watch` will watch for changes, and if any files in the `docs`
directory change, the `mix docs` task will be executed, regenerating the
documentation.

### DIFF
--- a/lib/mix/tasks/docs.watch.ex
+++ b/lib/mix/tasks/docs.watch.ex
@@ -1,0 +1,43 @@
+defmodule PhoenixGuides.Watcher do
+  use GenServer
+
+  def start_link() do
+    GenServer.start_link(__MODULE__, :ignored)
+  end
+
+  def init(_) do
+    :ok = Application.ensure_started(:fs)
+    :ok = :fs.subscribe()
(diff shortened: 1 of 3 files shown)


Gazler:
cc @chrismccord @lpil

lib/mix/tasks/docs.watch.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Watcher do
```
chrismccord: namespace?

lib/mix/tasks/docs.watch.ex
```diff
@@ -0,0 +1,42 @@
...
+  use GenServer
+
+  def start_link() do
+    GenServer.start_link(__MODULE__, :ignored)
+  end
+
+  def init(_) do
+    :ok = Application.ensure_started(:fs)
+    :ok = :fs.subscribe()
+    {:ok, File.cwd!() <> "/docs"}
```
chrismccord: let's make state a map, with `:path` as single key

chrismccord:
couple small notes, then :ship:

PR #624, opened 2016-11-13
Fix formatting of text blocks

rockwood:
For whatever reason \`\`\`text blocks revert to single-line mode. Changing  to \`\`\`console fixes it.

Also, mixed character types in directory trees causes the formatter  to revert to single-line mode (Apparently these are different: |│)


Gazler:
Great spot, thanks! <3

PR #626, opened 2016-11-24
Replacing empty map with :empty

ThrashAbaddon:
If no parameters are specified, we can’t just default to an empty map, because that
would be indistinguishable from a blank form submission. Instead, we default params
to the :empty atom. By convention, Ecto will produce an invalid changeset, with empty
parameters. - from Programming Phoenix

### DIFF
--- a/docs/ecto_models.md
+++ b/docs/ecto_models.md
@@ -224,7 +224,7 @@ defmodule HelloPhoenix.User do
   @doc """
   Builds a changeset based on the `struct` and `params`.
   """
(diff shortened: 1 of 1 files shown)


Gazler:
Thanks. While this was the case in Ecto 1.x (as used in Programming Phoenix), this is no longer the case.

As of Ecto 2.0, this is no longer the case. The relevant section of the changelog is https://github.com/elixir-ecto/ecto/blob/v2.0/CHANGELOG.md#revamped-changesets

PR #633, opened 2016-12-14
Modified url for highlighting correct lines

userimack:
Modified url for highlighting correct lines as the range given previously was not pointing the correct lines.


Gazler:
Thanks for this. If we link directly to the tag then we won't need to update this as often.

If you change it to https://github.com/elixir-lang/plug/blob/v1.3.0/lib/plug/conn/status.ex#L9-L69 then I can merge it in.

PR #640, opened 2017-01-02
Modified url for highlighting correct lines (amended)

EduardoBautista:
This addresses the comment left by @Gazler in https://github.com/phoenixframework/phoenix_guides/pull/633


Gazler:
Thanks <3

PR #643, opened 2017-01-09
Add some clarification why some static assets won't load

homanchou:
I also came across this confusion https://groups.google.com/forum/#!msg/phoenix-talk/jOKj8Myy5nc/Fj2Dhl_hGJMJ this little bit of documentation hopefully clarifies why some assets do not load unless included in :only option of Plug.Static


Gazler:
Thanks <3

PR #646, opened 2017-01-23
Postgres Installation: Note about default encoding

prabugp:
On OSX, when Postgres is installed using brew, the default encoding is set to SQL ASCII as explained https://www.postgresql.org/docs/current/static/app-initdb.html. This caused an error when doing `mix ecto.create` in the Up and Running section. Creating Postgres DB with the UTF-8 fixes this problem. 

I believe this "note" is needed because people who are new to Postgres and starting up with Elixir and Phoenix might think this is a Phoneix/Elixir/Erlang issue and not necessarily Postgres issue.


Gazler:
I don't use a mac. I'll find out if others have this issue and get back to you.

chrismccord:
We haven't seen the encoding issue int he wild so until we get more reports I think it's best to assume it relates to a quirk in your setup. Thanks for the PR!

PR #647, opened 2017-01-25
Update static_assets.md

shvaikalesh:


Gazler:
Could we keep the semantic versioning constraint of `>= 2.0` here?

shvaikalesh:
Until Brunch 3 is out, both are fine. However, `^2` is more future-proof.

chrismccord:
❤️❤️❤️🐥🔥

PR #653, opened 2017-02-14
Update installation.md to create "postgres" role

lvela:
I ran into this issue when going through the Installation guide and had to look up how to fix it. This should make it easier for beginners to get started. Found example in http://www.phoenixframework.org/v0.11.0/docs/mix-tasks


Gazler:
Thanks, however this is already covered in the guides at https://github.com/phoenixframework/phoenix_guides/blob/master/docs/bonus_guides/mix_tasks.md#ectocreate

We should probably link to the above from https://github.com/phoenixframework/phoenix_guides/blob/master/docs/ecto_models.md (where it mentions ecto.create). Feel free to open a new PR for that.

PR #656, opened 2017-02-21
more clear controllers.md

nhooyr:
The present example does not show the plug's welcome message being
overridden like it says so in the text. This new example more clearly
matches the text.


Gazler:
Thanks <3

PR #657, opened 2017-02-22
Remove HTML resource section from model test guide

krankin:
As I was reading through the testing models guide this section seemed out of place. I thought that moving it to the top of the guide might be appropriate but there's already a link to the Ecto models section that a reader could use for instructions on creating the sample User resource.

Since this section seems both out of place and redundant to me, I'm suggesting it be removed.


zzak:
:+1: I also thought this felt out of place.

Also here in the testing/intro guide:
https://github.com/phoenixframework/phoenix_guides/blob/master/docs/testing/testing.md#generating-more-files

This actually comes from the beginning of the Models guide:
https://github.com/phoenixframework/phoenix_guides/blob/41b4d8fa35b6ce932d49d6eaf2640632485f12b2/docs/ecto_models.md

Which I would assume people read _before_ getting to testing models..

Gazler:
The models guide states:

> Notice that we get a lot for free with this task - a migration, a controller, **a controller test, a model, a model test**, a view, and a number of templates.

Which I think is enough to convey that the generators also generate tests. It might be worth linking through to the testing pages in that paragraph?

chrismccord:
We are in the process of overhauling the guides, so unfortunately that means that some contributions will no longer be relevant. We appreciate you helping to improve the guides! Thanks!

PR #661, opened 2017-02-25
clarified misleading sentence in ecto_models.md

nhooyr:
cast/3's third argument is the allowed fields, not the required ones.

### DIFF
--- a/docs/ecto_models.md
+++ b/docs/ecto_models.md
@@ -248,7 +248,7 @@ Let's take a closer look at our default changeset.
   end
 ```
 Right now, we have two transformations in our pipeline to this model.
-In the first call, we invoke `cast/3` to send in our parameters and which fields are required for validation.
+In the first call, we invoke `cast/3` to send in our parameters and which fields are whitelisted for validation.
 `cast/3` first takes a struct, then the parameters (the proposed updates), and then the final field is the list of columns to be updated.
 `cast/3` also will only take fields that exist in the schema.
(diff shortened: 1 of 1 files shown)


digitalcake:
I'll need to double check, but I think the word "required" is correct in that I believe cast asks like a whitelist and that if its not in that list it will not be available in the corresponding pipe. 

IE: 
"All parameters that are not explicitly allowed are ignored."
https://hexdocs.pm/ecto/Ecto.Changeset.html#cast/3

nhooyr:
I don't think the word required gets that meaning across. I think `allowed` is more clear, even the function definition and documentation use allowed and not required. While cast acts like a whitelist, it allows only those on it's whitelist through, validate_required is still required to make sure that the required fields are present.

digitalcake:
why not the word "whitelisted" that an even more concrete terminology that I think gets the point across.

nhooyr:
Done

Gazler:
Sorry I'm late to this. I think we should go for `allowed` as that is what Ecto use in their docs https://hexdocs.pm/ecto/Ecto.Changeset.html#cast/3

nhooyr:
I'll try and update this soon.

nhooyr:
@chrismccord did you close due to inactivity? If so, I can update this now, I just forgot about it.

chrismccord:
Yes, inactivity, sorry. We are in the middle of a big overhaul on Ecto so this will be taken care of along with that effort. Thanks for the PR!

PR #664, opened 2017-03-06
update link to HTTP code mapping

jsallis:
Currently, the link to HTTP code mappings is highlighting incorrectly. This fix adjusts the highlighted line numbers.

### Before
before
### After
after


jsallis:
Looks like this is a dupe of #640. Closing in favor of that ticket.

PR #672, opened 2017-04-03
Update Configuration guide with inline example

sngeth:
Added a config/3 example to familiarize new users how to quickly add and
use use new environments variables and added a link to Mix.Config module.

### Summary
We should mention how configuration files work at a basic level or at least link to the Mix.Config module to show how the config function works.


chrismccord:
We are in the process of overhauling the guides, so unfortunately that means that some contributions will no longer be relevant. We appreciate you helping to improve the guides! Thanks!

PR #673, opened 2017-04-05
Fail fast if SECRET_KEY_BASE is missing

tsubery:
When it's missing, the compilation succeeds and you get very opaque error message. Better to fail sooner with a hint about the cause.


Gazler:
Thanks <3

PR #675, opened 2017-04-19
Phoenix 1.3

bratsche:
I know Phoenix 1.3 is backwards compatible, but I figured at some point the guides would want to be updated to use the new `mix phx.*` commands and the new directory layout. This is a beginning for that.

There's still plenty of stuff that needs updating, so I'll keep committing to this after work.


Gazler:
I merged this locally. Thanks <3

PR #676, opened 2017-05-04
Add note to Heroku deploy guide about static asset buildpack

lizvdk:
Mention that the entire "Adding The Phoenix Static Buildpack" section can be skipped for API only applications in the "Heroku" guide.

The app will fail to compile and deploy if it has the static buildpack without static assets.

This will be helpful for beginners.


Gazler:
Thanks <3

PR #678, opened 2017-05-14
Removing note for importing Router.Helpers

alvincrespo:
As I was running through the guide, as a first time user, I expected that I needed to add  `import HelloPhoenix.Router.Helpers` to the application view. But I didn't have to, so this line might be a bit confusing if you're following along.

### DIFF
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -184,7 +184,8 @@ This is significant because we can use the `page_path` function in a template to
 ```html
 <a href="<%= page_path(@conn, :index) %>">To the Welcome Page!</a>
 ```
-Please see the [View Guide](views.html) for more information.
+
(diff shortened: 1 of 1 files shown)


Gazler:
I agree this is confusing. I think we want to reference somewhere that the `import` is provided in the `web.ex` file. 

Perhaps something like:

> This is significant because we can use the `page_path` function in a template to link to the root of our application. We can then use this helper in our templates:
> ...
> The `page_path` function is imported into our template with `use HelloPhoenix.Web, :view`. Please see the [View Guide](views.html) for more information.

alvincrespo:
Thanks, this totally makes sense now. Ill make the update and push it up.

alvincrespo:
Ok just made the change. Thanks for the review! Cheers!

chrismccord:
❤️❤️❤️🐥🔥

Issue #680, opened 2017-05-15
Guides overhaul


chrismccord:
It's time to pay down my guides debt with a review and overhaul of the guides and much needed TLC. 

cc @Gazler, @willykaram, @jeregrine 

Here's a shortlist to get us 1.3.0 stable ready. I'm marking the ones that I'm claiming, but others are up for grab if folks have bandwidth on refreshing the existing guides:

* [x] Existing* guides in need of a refresh to include new features and latest conventions:
  * [x] [@heatherm] Installation
    * [x] needs to be bumped to 1.3 version requirements
      * [x] Erlang 18
      * [x] Elixir 1.5
    * [x] needs to reference new `phx_new` archive urls
    * [x] needs to use new `phx.new` task
  * [x] Up And Running
    * [x] needs to use new `phx.new`, `phx.server` tasks
    * [x] all example output needs to be updated to new output
  * [x] [CM] Adding Pages
    * [x] needs to discuss new directory structure and update tree listings
    * [x] needs to discuss new assets directory. `web/static` has moved to `assets`. `web/static/assets` has moved to `assets/static`, etc.
    * [x] modules need to use new `Web` namespace
    * [x] `Map.get(assigns, :messenger)` should be `assigns.messenger` since we've long raised on missing assigns
    * [x] __new additions__
      * [x] make small mention of EEx engine escaping by default when talking about templates (explained in detail later in Views guide, but should be mentioned here)
  * [x] [@gazler] Routing
    * [x] web namespace
    * [x] needs to use `phx.routes` with new example output
    * [x] __new additions__
      * [x] cover basic `forward` usecases and usage
  * [x] Plug
    * [x] null-arity calls should use parens
    * [x] web namespace in examples
  * [x] Controllers
    * [x] web namespaces
    * [x] revamp code examples to follow conventions, such as always using parens when piping, no single line pipe.
    * [x] __new additions__
      * [x] discuss `halt` when talking about controller plugs and its importance to prevent invoking the controller action
      * [x] action_fallback
  * [x] Views
    * [x] web namespaces
    * [x] update file path locations
    * [x] output of iex examples need refreshed
    * [x] __new additions__
      * [x] [CM] should show examples of functions on the view being available in eex templates and discuss why
  * [ ] Channels
    * [x] should cross-ref TBD presence guide at the end
    * [x] web namespace
    * [x] file paths
    * [ ] should warn about using `handle_out` with potential N query problems
[...]

PR #683, opened 2017-05-23
installation :: phoenix ~> phx

gemantzu:
Is it ok for me to continue doing some minor changes that I can help with? Or should I ask for permission to do a change in case someone has it already?


Gazler:
Thanks <3

Doing changes is fine, just keep an eye on https://github.com/phoenixframework/phoenix_guides/issues/680 - the list at the top should be updated if people claim a task.

gemantzu:
Yeah I just saw that Installation was taken. If @heatherm has already done some work on this I can dismiss it.

PR #686, opened 2017-05-23
Plug :: Null arity function calls

gemantzu:
Shouldn't we move the pipes to newlines as well?


chrismccord:
pipes can stay ❤️❤️❤️🐥🔥

gemantzu:
Ok 👍

PR #697, opened 2017-06-18
Fix Nginx config comment

denispeplin:
Comment on first line of Nginx config breaks Nginx if copied without changes.
It must start from `#`.


Gazler:
Thanks <3

PR #698, opened 2017-06-19
Update Template Guide for 1.3

HusseinMorsy:
Updates Templates guide for 1.3

* add web namespaces
* update file path locations

Maybe we should add todo items for the templates guide in #680.

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -4,7 +4,7 @@ Templates are what they sound like they should be: files into which we pass data
 
 EEx is the default template system in Phoenix, and it is quite similar to ERB in Ruby. It is actually part of Elixir itself, and Phoenix uses EEx templates to create files like the router and the main application view while generating a new application.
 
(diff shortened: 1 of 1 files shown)


willykaram:
@HusseinMorsy Thanks for helping with this. I gave this a quick peer review, and I think it looks good. My only suggestion would perhaps be including a line about the transition to 1.3 and the directory structure/module naming. However, I'll defer to @Gazler and the core team as to whether they want to include any of that in the individual guides, or keep that self-contained entirely in a 1.3 transition guide.

Gazler:
@willykaram That's a good suggestion, i think we should keep those sort of changes to the upgrade guides. Going forward, it will be easier to match the version of the guides against the version of the framework as the version specified on hexdocs will match. For example https://hexdocs.pm/phoenix/1.3.0-rc.2/overview.html

Gazler:
Thanks <3

PR #700, opened 2017-07-06
Recommend an easier install of Postgres for Mac users

arecvlohe:
I wanted to add an easier install of Postgres for Mac users with the Postgres App, a Postgres Mac GUI. This seemed to be the easiest way to get up and running with Postgres on my computer. I saw in other forums people were having dificulty with this as well, that's how I found Postgres App. Hopefully this can help others too.

I also fixed a type on line 92 where is said Postgrex.


Gazler:
Thanks for the suggestion and pull request!

I don't think we need this change. We already link to https://wiki.postgresql.org/wiki/Detailed_installation_guides which provides installation instructions (including Postgres App).

In this instance, "Postgrex" is correct as it is referencing the Elixir driver https://github.com/elixir-ecto/postgrex and not the database server itself.

PR #711, opened 2017-07-25
Use Elixir comment hash sign instead of C-like //

tompave:
Fix syntax issue in a code example that was confusing the syntax highlighter of [the HTML page](https://hexdocs.pm/phoenix/1.3.0-rc.3/seeding_data.html#content).


Gazler:
Thanks <3

PR #717, opened 2017-07-27
Shorten headers and sub-headers in context guide

Gazler:
Shorter headers are easier to read when the docs are published to hex
docs.

Old on the left, new on the right.

new-phoenix-headers


chrismccord:
:shipit:

PR #718, opened 2017-07-29
Fix image links

khaledh:
My first contrib, hope it goes well! 🤞

Images are currently broken on hexdocs.pm. The current links expect images to be at `/images/*`, which works when previewed on GitHub, but they are actually copied into `assets/*` by `ex_doc` when published, and so they're broken.

This PR moves images into `docs/assets/images/*` and updates the links to be relative: `assets/images/*`, so they would work both on GitHub preview as well as when published to hexdocs.pm.


Gazler:
Thanks <3

PR #719, opened 2017-07-29
Fix missing link to Ecto Section

sajmoon:
I don't know how to test this locally, if there is a way to generate the documentation locally, but I hope this change will link to https://hexdocs.pm/phoenix/ecto.html


Gazler:
Thanks <3

PR #720, opened 2017-07-29
Highlight the different folders

sajmoon:
The difference between hello and hello_web folder is clear when you
expect it to be in a different folder. But during my first read I missed
it and assumed it would be in _web.

The difference is important and should be pointed out more clearly.

The change is the addition of the sentence: "Note the difference between `lib/hello` and `lib/hello_web`."


Gazler:
Thanks <3

PR #722, opened 2017-07-30
Improve context guide

nekath:
While I was reading whole guide, I've tried to fix:
- update code snippets that comes from generator
- fix typos
- improve readability in few places

I'm not native speaker, so I've not check grammar.

### DIFF
--- a/docs/contexts.md
+++ b/docs/contexts.md
@@ -130,6 +130,7 @@ defmodule HelloWeb.UserController do
   use HelloWeb, :controller
 
   alias Hello.Accounts
+  alias Hello.Accounts.User
 
   def index(conn, _params) do
     users = Accounts.list_users()
@@ -137,7 +138,7 @@ defmodule HelloWeb.UserController do
   end
 
   def new(conn, _params) do
-    changeset = Accounts.change_user(%Hello.Accounts.User{})
+    changeset = Accounts.change_user(%User{})
     render(conn, "new.html", changeset: changeset)
   end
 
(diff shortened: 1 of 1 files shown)


docs/contexts.md
```diff
@@ -390,28 +392,33 @@ Next, let's show the user's email address in the user show template. Add the fol
 
 ```
 
-Now if we visit `http://localhost:4000/users/new`, we'll see the new email input, but if you try to save a user, you'll find that the email field is ignored. No validations are run telling you it was blank and the data was not saved. What gives?
+Now if we visit `http://localhost:4000/users/new`, we'll see the new email input, but if you try to save a user, you'll find that the email field is ignored. No validations are run telling you it was blank and the data was not saved, and at the end you'll get an exception `(UndefinedFunctionError) function nil.email/0 is undefined or private`. What gives?
```
nekath: Imho, It's helpful for beginners, to note, that exception will occur.

docs/contexts.md
```diff
@@ -875,18 +882,18 @@ With our new plugs in place, we can now modify our `create`, `update`, and `dele
     end
   end
 
-  def update(conn, %{"page" => page_params}) do
++ def update(conn, %{"page" => page_params}) do
```
jeregrine: This doesn't require a `+` because it was already there?

nekath: We should use `plus` signs to mark all changes that are need to be done in the CMS.PageController.

In this line, the `update` function is generated as `def update(conn, %{"id" => id, "page" => page_params})`. So we need to remove `id` and because of that the `plus` sign was added

nekath:
@jeregrine I've rebased and fixed conflicts.

nekath:
I've rebase my branch again, and removed questionable change in page controller.

jeregrine:
Sorry for delay @nekath thank you for your help <3

nekath:
@jeregrine No problem :)

PR #725, opened 2017-07-31
Fix links in context guide

RudolfHattenkofer:


chrismccord:
Thank you! Can you make the guide links local file references so they work for both local docs builds and hexdocs?  Thanks!

PR #728, opened 2017-08-01
Updated web.ex file to reflect new directory structure in template guide.

kunashu:
* Updated the reference to web.ex to hello_web.ex to reflect the new structure change in the phoenix project.

* Also updated the view imports to reflect how it looks in a vanilla phoenix project.


Gazler:
Thanks <3

PR #733, opened 2017-08-03
Fix GenStage link in Running Asynchronous Tasks

alejandrodnm:
The link to GenStage is broken since it's a reference to when GenStage was experimental


Gazler:
Good spot. Thanks <3

PR #740, opened 2017-08-06
Update Phoenix's deprecated task names

franckverrot:
This commit makes sure users don't get any of the

> `mix phoenix.<task_name>` is deprecated. Use `phx.<task_name>` instead."

messages.


Gazler:
Thanks <3

PR #741, opened 2017-08-06
Clarify endpoint concept

jwogrady:
As a new Phoenix developer I needed to understand that "endpoint" is the boundary that separates external from internal requests. The endpoint isn't, as the name might imply, where requests terminate. It's called "endpoint" because that's as far as the external requester can go before it has to wait for the Phoenix app to process the request and respond. From the Phoenix developer's point of view the endpoint is more like the beginning than the end!

### DIFF
--- a/docs/introduction/overview.md
+++ b/docs/introduction/overview.md
@@ -15,7 +15,7 @@ Phoenix is actually the top layer of a multi-layer system designed to be modular
 Phoenix is made up of a number of distinct parts, each with its own purpose and role to play in building a web application. We will cover them all in depth throughout these guides, but here's a quick breakdown.
 
  - [Endpoint](endpoint.html)
-    - where external requests are received and internal processing begins
+    - the start and end of the request lifecycle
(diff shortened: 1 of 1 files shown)


Gazler:
Thanks for opening this! I agree with the concept of the PR, I'm just not 100% on the concept of internal/external, I'll get back to you on this one.

docs/introduction/overview.md
```diff
@@ -15,6 +15,7 @@ Phoenix is actually the top layer of a multi-layer system designed to be modular
 Phoenix is made up of a number of distinct parts, each with its own purpose and role to play in building a web application. We will cover them all in depth throughout these guides, but here's a quick breakdown.
 
  - [Endpoint](endpoint.html)
+    - where external requests are received and internal processing begins
```
jeregrine: I think I would prefer something discussing that this is the application entry point. All connections start here. "Thinking of it as the original pipeline"

Gazler:
Thanks <3

PR #742, opened 2017-08-06
change bullet order under Views overview

jwogrady:
Reads more naturally and helps emphasize the order of operation.


Gazler:
Thanks for the PR. I'm not sure this particular change is required. I actually "Views render templates" to "Views render the templates".

PR #744, opened 2017-08-06
format list of actions under Controllers overview

jwogrady:
Made action lower case for format consistency, and added a colon to emphasize that bullet is functioning as a list heading.


Gazler:
Thanks <3

PR #755, opened 2017-08-09
Fixing various formatting issues or typo

weppos:
First of all, thanks for the effort you put writing these guides. They are an incredibly helpful resource.

I took note of some issues I found while going through the various pages. I packaged most of them into a single PR, instead of a creating a ticket for each one that would result in a lot of noise.

I left the single changes as separate commits, so that you can easily tell me if I should rollback and discard any of them.


Gazler:
Can you leave out the following files from the HTTPS change?

 * styling/Phoenix.html
 * styling/Phoenix_files/_hero.scss

I think we will aim to remove the "styling" directory in the future now that there is a phoenix_site repo.

Gazler:
Thanks! All good except from that 1 change. I'll merge after that.

weppos:
Sure thing! I rebased the list, and removed the https change completely. I will propose it as different PR, without the changes to style.

I though it would be the best way, as reverting that patch would have still touched the files in history (and potentially caused conflicts when you'll remove the assets). Also, editing that specific commit was challenging as well (as it was the first).

The changeset now includes only the other 8 commits.

Gazler:
@weppos Sorry, it looks like another PR I have merged has resulted in some conflicts, would you mind fixing them up please?

weppos:
Done. It looks like the conflicting commit was 707ae2ed06e41ee37b78b2b1168167115703421f

Gazler:
Thanks <3

PR #759, opened 2017-08-16
Update up_and_running.md, for better explain 

TangMonk:
https://github.com/phoenixframework/phoenix/issues/2435

### DIFF
--- a/docs/up_and_running.md
+++ b/docs/up_and_running.md
@@ -93,6 +93,18 @@ Then configure your database in config/dev.exs and run:
 
     $ mix ecto.create
 
+In ./config/dev.exs, You should properly fill the database configuration
+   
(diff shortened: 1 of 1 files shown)


Gazler:
Thanks for the PR. Could we change this by modifying the following line instead of adding the entire config block that is already available in `config/dev.exs`

Instead of:

    Now we'll create our database:

We could have:

    Now we'll create our database (please ensure the configuration in `config/dev.exs` matches your local database credentials):

Gazler:
Phoenix already prompts you to configure the database when running `mix phx.new`

    Then configure your database in config/dev.exs and run:

TangMonk:
Sorry...,  I am not read it carefully

PR #760, opened 2017-08-17
Remove MIX_ENV=prod from mix phx.digest commands

jayjun:
`MIX_ENV` is not used by `mix phx.digest` in any way.


jeregrine:
Thanks for this! I did some research to confirm and it appears the only place we reference the config is https://github.com/phoenixframework/phoenix/blob/2e911762d672fa1be2868e02d07745cc91bc7291/lib/phoenix/digester.ex#L205 and that's only overridable if you know what you are doing. 

I only mention that for historic context.

Thanks! @jayjun 💟

PR #764, opened 2017-08-19
Fix 'TODO'

tomconroy:
I believe the only file added from the Ecto guide is this one


tomconroy:
hm, this duplicates #762 but with another fix

jeregrine:
@tomconroy Thanks for the contribution but this is indeed fixed! <3

PR #765, opened 2017-08-19
Fetch session before put session

user-tony:
Detail: https://github.com/phoenixframework/phoenix/issues/173


user-tony:
unnecessary update, close it now.

PR #769, opened 2017-08-23
Fix module call Endpoint as HelloWeb.Endpoint after call 'iex -S mix'

ramortegui:
On the guide we are starting a new iex session, and we aren't importing HelloWeb.Router.Helpers  nor aliasing HelloWeb.Endpoint.  so we need to call the Endpoint with a full name.


Gazler:
Thanks <3

PR #770, opened 2017-08-23
Add Plug.RequestId to the defaults Plugs at Endpoint plugs section on…

ramortegui:
On Phoenix 1.3.0 RequestId plug was added by default at hello_we/endpoint.ex.  The commit will add the info about it.


Gazler:
Thanks <3 This has been in Phoenix for a long time, must have been missed when the list was originally created in the guide. https://github.com/phoenixframework/phoenix/blob/v1.0/installer/templates/new/lib/application_name/endpoint.ex

PR #773, opened 2017-08-23
Prefer https over http when possible

weppos:
This is a follow up to #755 where I was asked by @Gazler to avoid changing the `styling` folder.

The switch to https is a trivial change compared to other major fixes, hence no very high priority here. It should be an easy merge though.

### DIFF
--- a/docs/adding_pages.md
+++ b/docs/adding_pages.md
@@ -249,7 +249,7 @@ def show(conn, %{"messenger" => messenger} = params) do
 end
 ```
 
-It's good to remember that the keys to the `params` map will always be strings, and that the equals sign does not represent assignment, but is instead a [pattern match](http://elixir-lang.org/getting-started/pattern-matching.html) assertion.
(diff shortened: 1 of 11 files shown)


Gazler:
@weppos To save doing these changes twice, would you be able to change all the docs to point to hexdocs?

For now let's change any references to:

https://elixir-lang.org/docs/stable

to:

https://hexdocs.pm/elixir/1.5.1

See this commit for inspiration https://github.com/elixir-lang/elixir/commit/fa8feaba59862de7526499a9177efd8b2ca09083

We should maybe find a way to make the version dynamic in the future so that we don't have the version hardcoded, but pointing to 1.5.1 will be fine for now.

jeregrine:
@Gazler Lets break that into a new task for tomorrow.

@weppos If you could fix merge conflicts we can work on getting this merged .

Gazler:
Thanks <3 <3

weppos:
Thanks for merging it. Sorry, I was not able to apply the requested changes due to other priorities in these weeks.

PR #776, opened 2017-08-29
Move the alias created by the generator.

ramortegui:
In order to make Credential.changeset function available on users functions the alias for Hello.Accounts.Credentials needs to be defined in advance.

This pull request adds the update of the alias on Hello/Accounts.ex.


Gazler:
Merged locally. Thanks <3

PR #777, opened 2017-08-29
Update missing instructions to update the delete function on CMS cont…

ramortegui:
The original delete function had:
def delete(conn, %{"id" => id}) do
  page = CMS.get_page!(id)
And page nor id aren't used to as we are retrieving the page from the assigns.
This push only update the code, and prevent warnings on the debug.


chrismccord:
❤️❤️❤️🐥🔥

PR #784, opened 2017-09-02
Contexts - code filled

jung-hunsoo:
On undoing previous User contexts, no actual code was provided. This fix should work.


jung-hunsoo:
Closed since already fixed in #762

PR #816, opened 2017-09-07
Use String.duplicate/2 instead.

jung-hunsoo:
An Elixir's String.duplicate/2 function can be used at no risk instead of creating our own function.


Gazler:
This is a great PR, however the guides have moved now. Please re-open the issue on this file in Phoenix https://github.com/phoenixframework/phoenix/blob/master/guides/docs/testing/testing_schemas.md

Thanks <3

